PR #10, opened 2021-08-30
remove / and * operators from function calls to support 3.6 and 3.7 #…

### LINKED ISSUE #9: failing build on 3.6 and 3.7 due to / in arguments list
I suggest to stop using this technique. Alternatively, we could stop supporting 3.6 and 3.7. I think we want to support these versions as we hope to build some tool that can be widely used. We might therefore not want to leave out people somehow bound to slightly older python versions.

cwmeijer:
…Fixes #9

### DIFF
--- a/dianna/__init__.py
+++ b/dianna/__init__.py
@@ -10,7 +10,7 @@ __email__ = "dianna-ai@esciencecenter.nl"
 __version__ = "0.1.0"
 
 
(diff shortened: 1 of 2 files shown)


loostrum:
Looks good to me! Minor point: the * operator was introduced in 3.0 so we could keep it, but I'm also fine with removing them both.

loostrum (approved):

PR #21, opened 2021-10-13
make rise and lime apis consistent

cwmeijer:
fixes 'https://github.com/dianna-ai/dianna-exploration/issues/138'
Introduces a `labels` parameter for both rise and and lime. Both return an object of the same structure.

### DIFF
--- a/dianna/methods/rise.py
+++ b/dianna/methods/rise.py
@@ -4,6 +4,10 @@ from tqdm import tqdm
 from dianna.utils import get_function
 
 
+def normalize(saliency, n_masks, p_keep):
+    return saliency / n_masks / p_keep
+
+
 class RISE:
     """
     RISE implementation based on https://github.com/eclique/RISE/blob/master/Easy_start.ipynb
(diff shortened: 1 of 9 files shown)


dianna/methods/rise.py
```diff
@@ -25,34 +29,44 @@ def __init__(self, n_masks=1000, feature_res=8, p_keep=0.5, preprocess_function=
...
+        self.masks = self._generate_masks_for_text(text_length)  # Expose masks for to make user inspection possible
+        sentences = self._create_masked_sentences(input_tokens)
+        saliencies = self._get_saliencies(runner, sentences, text_length, batch_size)
+        return self._reshape_result(input_tokens, labels, saliencies)
+
+    @staticmethod
+    def _reshape_result(input_tokens, labels, saliencies):
+        word_lengths = [len(t) for t in input_tokens]
+        word_indices = [sum(word_lengths[:i]) + i for i in range(len(input_tokens))]
+        return [list(zip(input_tokens, word_indices, saliencies[label])) for label in labels]
```
egpbos, line 45:
```suggestion
        return list(zip(input_tokens, word_indices, *[saliencies[label] for label in labels]))
```

egpbos: Never mind :P Now I understand the LIME output.

PR #50, opened 2021-10-25
Run tests also on python 3.10

### LINKED ISSUE #49: Support python 3.10
3.10 is now a full release and we should support it.

loostrum:
Closes #49

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.7', '3.8', '3.9']
+        python-version: ['3.7', '3.8', '3.9', '3.10']
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/install-python-and-package
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,7 @@ classifiers =
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
 description = Deep Insight And Neural Network Analysis
 keywords =
     XAI


egpbos:
This is blocked on onnxruntime releasing a pypi package for 3.10, see https://github.com/microsoft/onnxruntime/issues/9782.

elboyran:
Are we still considering 3.10?

loostrum:
Patrick's comment still applies I think. We should support 3.10 when we can (which just means running the tests, there's nothing breaking in 3.10)

egpbos:
FYI, I rebased on main.

egpbos:
onnxruntime 1.12 is out and it has a 3.10 pypi release! I rebased this branch on main, hope the tests pass now 🤞

egpbos:
Ok, the whole thing seems to pass now, but still have to remove three commits that are no longer necessary, will rebase now...

egpbos:
I have no idea why linting fails now, because I didn't change any linted code. I think it should also be an issue on the main branch right now, but haven't seen it there.

elboyran:
> I have no idea why linting fails now, because I didn't change any linted code. I think it should also be an issue on the main branch right now, but haven't seen it there.

I have also added a sentence at the Tutorials readme and the linting failed :-(.

egpbos:
Ok, pylint uses astroid and astroid crashes (for some reason) with a recursion depth error. This error was fixed in astroid 2.12.2 (the most recent version), but pylint has set an upper limit on its astroid requirement with `astroid<=2.12.0-dev0,>=2.11.6`, so it won't install the right version.

I manually installed the more recent astroid version on my laptop and I see no linting errors (except actually a new one that is only present for Python 3.10, but unrelated to this PR... let's fix that when we switch fully to 3.10 for CI default). So I think we can safely merge this now.

We could in theory add a manual install step to install the latest astroid manually on CI, but I think it would be better to just wait for the upstream fix and be careful in the meantime (i.e.: check linting on laptop).

So, in short: ready to merge :)

egpbos:
Ah, yeah, we just have to wait for the next pylint version for this issue to resolve itself, the dependency was already updated in pylint: https://github.com/PyCQA/pylint/pull/7153

egpbos:
@elboyran once you give an approving review, the PR will be merged.

elboyran (approved):

elboyran:
@egpbos how to update the badge of supported versions?

egpbos:
That should go automatically after a new release. The badge shows versions on PyPI.

PR #61, opened 2021-11-16
Remove captum-based deeplift implementation; Fixes #60

### LINKED ISSUE #60: Remove DeepLift implementation
And its dependency on captum

loostrum:


cwmeijer (approved):
It fails half of the tests with obscure installation problems. Looking at your changes however, it all seems fine, simple and as expected. I don't think the failing tests have anything to do with your changes.
Merge it 👍

egpbos:
I can quickly fix the install errors before merging.

egpbos:
Lemme also quickly fix the linting failure...

egpbos:
LGTM now! ✅ 

At the risk of becoming the annoying holier than thou pointing-fingers OCD guy of the team: let's try to get green CI lights on every PR ;) The 🔴 from the linting check *was* actually caused by changes in this PR. Nothing critical in this case, but just goes to show that it's good practice to keep them green always.

loostrum:
Thanks both! I agree green lights are good for every PR. Too bad that the linting error was hidden by the install failure, but with requiring everything to pass before merging we'd still catch issues like that.

loostrum:
I've also removed the torch dependency as we don't use it anymore. Assuming you agree with this minor change I'll merge it when all lights turn green :)

PR #73, opened 2021-12-14
Allow all method keyword arguments

### LINKED ISSUE #68: LIME explain_image design choice; forward kwargs to methods
LIME `explain_image` directly returns 
`mask = explanation.get_image_and_mask(label, positive_only=False, hide_rest=True, num_features=num_features)[1]`
without giving the user the opportunity to change `positive_only`, `hide_rest`, or play with parameters such as `min_weight`.

What was the reasoning behind this choice of default parameters? 
Would it be useful to give the user the opportunity to pass those values if needed?

The user will likely stumble upon the base LIME implementation documentation even when trying to use the DIANNA wrapper and that documentation demonstrates interesting flexibility in how to visualize the mask:
https://marcotcr.github.io/lime/tutorials/Tutorial%20-%20images.html

egpbos:
This PR will allow users to pass along all possible keyword arguments to the explainer methods. It does this by adding `**kwargs` to the `explain_image/text` functions and passing them along appropriately (i.e. forwarding the proper keyword arguments to the proper call sites; some explanation methods need multiple calls and these calls may not support the same set of keyword arguments, so we need to filter the `**kwargs` dict for which a helper function was created in `utils`).

I propose to also remove a few existing keyword arguments that had the same defaults as those in the explainer libraries. These can now still be changed by the user by providing it through `**kwargs`, but we don't have to maintain the default values anymore.

Finally, I update documentation on our side a bit. We listed all parameters, but didn't explain them. I now simply refer to the method's documentation for those.

~This is still WIP, I only did LIME for now.~ LIME is also the only method to which this is relevant, right?

Oh, and ~another~ two other things: The first commit in this PR "fixes" testing and coverage configuration. Maybe this breaks CI in some way, in which case I'll try to fix it. But the reason is that the previous configuration broke test discovery in IDEs (VSCode and PyCharm at least). This should fix that. Please try it out! The second thing is a small commit that adds `tox` support. Now that configuration is all worked out, `tox` is a super convenient way to run tests on your laptop in a clean environment. This helps catch dependency configuration errors a bit faster, because you may have installed deps in your local environment manually but forgot to add them to the deps list. Just personal preference, using pytest without tox can work fine as well :)

Fixes #68

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Check style against standards using prospector
         run: prospector --zero-exit --output-format grouped --output-format pylint:pylint-report.txt
       - name: Run unit tests with coverage
-        run: pytest --cov --cov-report term --cov-report xml --junitxml=xunit-result.xml tests/
+        run: pytest --cov --cov-report term --cov-report xml --cov-report html --junitxml=xunit-result.xml tests/
       - name: Correct coverage paths
         run: sed -i "s+$PWD/++g" coverage.xml
       - name: SonarCloud Scan
--- a/.gitignore
+++ b/.gitignore
@@ -8,10 +8,13 @@ dist
 .cache
 __pycache__
 
+# testing
 htmlcov
 .coverage
 coverage.xml
 .pytest_cache
+.tox
+word_vectors.txt.pt
 
 docs/_build
 
--- a/dianna/methods/lime.py
+++ b/dianna/methods/lime.py
@@ -1,6 +1,6 @@
 from lime.lime_image import LimeImageExplainer
 from lime.lime_text import LimeTextExplainer
-from dianna.utils import get_function
+from dianna import utils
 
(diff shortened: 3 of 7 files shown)


egpbos:
Do we also need tests for this?

egpbos:
Ready for review!

loostrum (approved):
Looks good to me! Nice helper function, and I like the explicit mention of the LIME docs in our docstrings so the user knows where to go. Is it possible to put URLs in a docstring that get rendered properly? That might be a nice addition.

egpbos:
Good idea @loostrum, lemme try adding links, then I'll merge after that.

egpbos:
I just added URLs. Not the nicest way, but I couldn't figure out anything better. There's also some kind of magic Sphinx syntax, but that doesn't produce URLs. I guess you could get that to work, but then the docstring itself is only well readable after processing with Sphinx, so that's not so nice for interactive use. So this is probably the best we can get.

PR #74, opened 2021-12-15
Lint docstrings as well

egpbos:
This commit adds docstring linting in prospector. There's plenty to fix, so we should probably do that in this PR as well before merging.

I also added `--diff` to `isort`, so that it now tells you how to fix things when it fails.


egpbos:
Made obsolete by #87

cwmeijer:
We changes some settings that makes linting more strict. This resulted in 85 prospector messages. We brought it down to 83. The plan is to mark this draft 'ready for review' when all checks pass again. This should be done fairly quickly in order to avoid merge conflicts as we are touching (almost) all files containing code.

cwmeijer:
Now it's 'just' 50 messages.

egpbos (commented):
LGTM (I cannot approve, since I am the PR author :D even though @cwmeijer did most of the work)!

I expanded the `visualization.plot_image` docstring, so maybe you could have a final look at that @cwmeijer or someone else, but maybe we can also just merge now and look at it later, because in general we did not think very critically about the contents of the docstrings here, this PR is just to activate linting on docstrings and make sure it doesn't give more linting errors. We already opened #107 to think about docstring contents more carefully.

cwmeijer:
2 pairs of eyes is enough. I'm merging it.

PR #75, opened 2021-12-15
Remove python 3.6 from setup.cfg and build matrix

### LINKED ISSUE #70: Remove Python 3.6 support
3.6 is end of life on 23 Dec 2021, after which it will no longer receive security updates.

egpbos:
Fixes #70.


loostrum (approved):
Looks good!

loostrum:
Just now I noticed 3.6 was also mentioned in project_setup, I removed it there.

egpbos:
Thanks for the review! Btw, I'm not sure we need to keep project_setup.md, but that's a separate cleanup issue :)

PR #76, opened 2021-12-16
Fix shape of output in rise test; closes #62

### LINKED ISSUE #62: Wrong shape in RISE test
`test_rise_filename` uses the MNIST model, but doesn't take into account that it needs a swap of the channels axis. (Model has it as 2nd axis, RISE wants it to be the last axis).
Now the heatmaps produced have a shape of `(1, 28)` instead of `(28, 28)`. The test passes, because it does match the input shape and that is the only thing we test for.
The test could use the `preprocessing_function` option to reshape the data before running the test, and perhaps explicitly test if the shape matches the (y, x) shape of the input.

loostrum:
The tests now correctly generates heatmaps of shape (28, 28), instead of (1, 28)

### DIFF
--- a/tests/test_rise.py
+++ b/tests/test_rise.py
@@ -6,6 +6,10 @@ from tests.utils import ModelRunner, run_model
 from .test_onnx_runner import generate_data
 
 
+def make_channels_first(data):
+    return data.transpose((0, 3, 1, 2))
+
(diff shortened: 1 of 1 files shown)


egpbos (approved):
Ship it!

Issue #59, opened 2021-11-04
Keep track of data types


loostrum:
In some cases, the user may provide data with one data type, but the XAI methods produces model input with another data type. We should keep track of this, and convert the data back to the correct type before feeding it to the model.

cwmeijer:
We don't fully understand what you mean. Could you give more details and a plan how to solve it?

loostrum:
In some cases the data/model it may e.g. expect float32, which somewhere in the XAI method gets converted to float64 and we'd have to convert back to float32 before feeding data to the model. 
Could be solved by storing the input dtype, then calling data.astype(dtype) before feeding to the model. I'll work on implementing that now

PR #79, opened 2021-12-16
Remove explicit dtype conversion in SimpleModelRunner

loostrum:
Now that we keep track of the dtype of the data given by the user, we no longer 
need to do any conversion explicitly. 

This PR removes the explicit conversion to float32 in SimpleModelRunner, and adds explicit conversion
of the input data used in tests.

### DIFF
--- a/dianna/utils/onnx_runner.py
+++ b/dianna/utils/onnx_runner.py
@@ -1,4 +1,3 @@
-import numpy as np
 import onnxruntime as ort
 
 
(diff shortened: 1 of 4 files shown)


loostrum:
Somewhat related to #59

egpbos:
Also followup of #77, right?

egpbos (approved):
It's a bit difficult to envision all the reprecussions all this type mangling may have for users, but that is something that we have no tests for at this point, since we have no actual usecases of when it is a problem (if it ever will be a problem at all).

So, I think for now this is fine to merge :)

egpbos:
There's a merge conflict now.

Issue #85, opened 2021-12-23
Use type hinting with mypy


cwmeijer:
It is getting a bit convoluted what the input types of specific functions are. We now have options: numpy array, xarray, string, list of strings. I think it would help both development and users to specify at certain places what types we expect.

egpbos:
Indeed, probably all the places where we currently feel that we have to specify it in the docstring (input arguments like `matrix (np.array): blabla`) we should have it as a type hint instead.

elboyran:
Seems important as we are adding soon more modalities.

stefsmeets:
We discussed this today in the retrospective. We agreed this is a good idea to add gradually.

My suggestion would be to just enable mypy and then see how it goes.

@geek-yang @cwmeijer

PR #87, opened 2021-12-23
Set prospector to strict linting

### LINKED ISSUE #84: Linting configuration changes between laptop and CI
As noted in the [discussion here](https://github.com/dianna-ai/dianna/commit/5684ac0d7ad6d594d6daece0cb24b0f1ea39a7ed), it seems that on local machines both linting (prospector) and isort yield different results than on CI. Specifically, CI runs seem to be more lenient than local runs, which means CI misses issues. We should find out why this happens and how to fix it.

egpbos:
Line length 127 is the average of 3*42 and 128, which are both nice numbers.

This makes #74 obsolete, because docstring linting is also included here now. ~Except that that branch changed the `isort` call, that should be added here as well.~

This may also fix #84, but we should make sure.

I also added mypy linting, in anticipation of #85. Mypy passes without messages right now.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -33,4 +33,4 @@ jobs:
       - name: Check style against standards using prospector
         run: prospector
       - name: Check import order
-        run: isort --recursive --check-only dianna
+        run: isort --check-only dianna --diff
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -4,10 +4,11 @@
 
 output-format: grouped
 
-strictness: medium
-doc-warnings: false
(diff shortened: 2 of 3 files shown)


cwmeijer:
What's the status and purpose of this PR right now? I hope you're ok with me changing it to draft for now.

egpbos:
The purpose is to keep around the changes made here so we can reference them later. Indeed: draft. Maybe we can close it later if everything has been merged elsewhere.

cwmeijer:
What do you mean by 'when everything has been merged elsewhere'? What changes are we waiting for?

PR #90, opened 2021-12-24
Add testing of tutorial notebooks using nbmake

### LINKED ISSUE #89: Test notebooks in CI
In today's review, it was noted that we do not currently check whether the notebooks still work after changes. In fact, they did not. We should see whether we can add tests for notebooks to CI.

I haven't looked into it yet, but perhaps it is as easy as just running the notebooks from the command line with [`nbterm`](https://github.com/davidbrochart/nbterm) and checking whether they produce exceptions or other errors.

egpbos:
Fixes #89.

### DIFF
--- a/.github/actions/install-python-and-package/action.yml
+++ b/.github/actions/install-python-and-package/action.yml
@@ -1,0 +1,24 @@
+name: "Install Python and DIANNA"
+description: "Installs Python, updates pip and installs DIANNA together with its dependencies."
+inputs:
+  python-version:
+    required: false
+    description: "The Python version to use. Specify major and minor version, e.g. '3.9'."
+    default: "3.9"
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Python ${{ inputs.python-version }}
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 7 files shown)


cwmeijer:
As it's still failing tests and there is no recent activity, we don't think you want us to review it, so I changed it to draft. ♥

egpbos:
Ok, reviewers, this PR now does three things:

1. Add nbmake jobs, which, obviously, fail currently. I opted not to fix them here, but just let them fail, since people are still working on the notebooks in other PRs as well. This is also the reason I made them separate jobs, instead of including them in the build jobs; build then still passes, as it should.
2. Use a new Actions feature called Composite Actions to remove duplication from the jobs. I added a custom Action that installs Python, DIANNA and dependencies. This action is then referenced from all the other jobs.
3. The new Action also has an added feature: it caches pip. This could be faster than reinstalling every time. It could also be slower :D This is just an experiment.

cwmeijer (approved):
Tested nbmake command locally. looks ok. Great addition.
Refactoring makes sense 👍
Thanks for the renaming of 'build' :-)
lgtm+

PR #93, opened 2022-01-06
92 kernelshap with images

### LINKED ISSUE #92: Implement KernelSHAP for use with images
Based on the exploration https://github.com/dianna-ai/dianna-exploration/blob/main/example_data/xai_method_study/SHAP/KernelSHAP_shap_mnist_binary_onnx2tf.ipynb

Use https://github.com/slundberg/shap

Our example shows how to work with onnx model through converting to tf model.

geek-yang:
Add KernalSHAP method to dianna to support images explanation.

A tutorial is also included.

close #92


geek-yang:
Note that this module does not support the high level api of dianna (e.g. dianna.explain_image) for now since it is not clear which keyword arguments should be shared by the image explainer and the text explainer. We can improve this after implementing this method to text data.

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+        https://github.com/slundberg/shap/blob/master/shap/explainers/_kernel.py
+                                 and the documentation of image segmentation via:
+        https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.slic
+
+        Returns:
+            Explanation heatmap of shapley values for each class (np.ndarray).
         """
-        Example call function.
+        self.onnx_model = onnx.load(model)  # load onnx model
+        self.output_node = prepare(self.onnx_model, gen_tensor_dict=True).outputs[0]
```
loostrum, line 120: Would it make sense to move the onnx to tf conversion to dianna.utils?

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+            segmentation (np.ndarray): Image segmentations generated by
+                                       the function _segment_image
+            image (np.ndarray): Image to be explained
+            background (int): Background color for the masked image
+            channel_axis (int): See the function explain_image
         """
-        self.model = model
+        # if the image shape is [color_channels, height, width]
+        if channel_axis_first:
+            image = np.transpose(image, (1, 2, 0))
```
loostrum, line 188: We should probably use the xarray tools here (and in other spots where the transpose is done) like we do for LIME, see here: https://github.com/dianna-ai/dianna/blob/88bcaec001e640c35e5e1e08517ef1624fd661cb/dianna/methods/lime.py#L159

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+        if background is None:
+            background = image.mean((0, 1))
+
+        # Create an empty 4D array
+        out = np.zeros(
+            (features.shape[0], image.shape[0], image.shape[1], image.shape[2])
+        )
+
+        for i in range(features.shape[0]):
+            out[i, :, :, :] = image
```
loostrum, line 199: Small suggestion that I think looks a little bit nicer: `out[i, ...] = image` would also work. Actually just `out[i] = image` works too.

loostrum: Thinking about it, perhaps the out array can be generated with something like `np.tile`. Then we may not need the double for loop anymore, but could do everything with np.where etc.

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+            for j in range(features.shape[1]):
+                if features[i, j] == 0:
+                    out[i][segmentation == j, :] = background
+        if channel_axis_first:
+            out = np.transpose(out, (0, 3, 1, 2))
 
-        return np.zeros(input_data.shape, dtype=float)
+        return out.astype(np.float32)
```
loostrum, line 206: Is the float conversion to make the model accept the modified data? Then probably it's better to not convert to float32, but to whatever the dtype of the original input data is.

loostrum (changes requested):
Nice work! I've added some comments to make it more consistent with the utils we use in the other methods.
You're adding two models in the test_data folder, but I don't think those are actually used anywhere?

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
 import numpy as np
+import shap
+import onnx
+from skimage.segmentation import slic
```
egpbos, line 4:
```suggestion
import skimage.segmentation
```

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+            https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.slic
         """
-        self.shap_arg_1 = shap_arg_1
-        self.shap_arg_2 = shap_arg_2
 
-    def explain_image(self, model, input_data):
+        if channel_axis_first:
+            image = np.transpose(image, (1, 2, 0))
+
+        image_segments = slic(
```
egpbos, line 52:
```suggestion
        image_segments = skimage.segmentation.slic(
```
This makes it easier to read for people that don't know `skimage.segmentation` by heart.

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
     """
-    def __init__(self, shap_arg_1=0, shap_arg_2=0):
+
+    def __init__(self):
+        """KernelSHAP initializer.
+
         """
-        Example constructor for testing.
+
+    def _segment_image(
```
egpbos, line 19: This function doesn't use `self`, so it can be moved outside of the class as a generic helper function.

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+        sigma,
+        spacing,
+        convert2lab,
+        enforce_connectivity,
+        min_size_factor,
+        max_size_factor,
+        slic_zero,
+        start_label,
+        mask,
+        channel_axis_first,
```
egpbos, line 34: To simplify this function, just replace all these explicit parameters with `channels_axis_first, **kwargs` and pass everything on to `slic` below as `skimage.segmentation.slic(image=image, **kwargs)` after popping `image` from `kwargs` to handle it separately.

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+        # other keyword arguments for the method segment_image
+        max_num_iter = kwargs.get("max_num_iter", 10)
+        spacing = kwargs.get("spacing", None)
+        convert2lab = kwargs.get("convert2lab", None)
+        enforce_connectivity = kwargs.get("enforce_connectivity", True)
+        min_size_factor = kwargs.get("min_size_factor", 0.5)
+        max_size_factor = kwargs.get("max_size_factor", 3)
+        slic_zero = kwargs.get("slic_zero", False)
+        start_label = kwargs.get("start_label", 1)
+        mask = kwargs.get("mask", None)
```
egpbos, line 133: All these are unnecessary if you just pass `kwargs`. The default values are also the same as `slic`'s, right? So those too can be left out in DIANNA. To determine which kwargs can be passed to `segment_image` (or rather `slic`), you can use `slic_kwargs = dianna.utils.get_kwargs_applicable_to_function(slic, kwargs)` (see here: https://github.com/dianna-ai/dianna/blob/main/dianna/utils/misc.py#L26).

dianna/methods/shap.py
```diff
@@ -1,21 +1,223 @@
...
+            max_num_iter,
+            sigma,
+            spacing,
+            convert2lab,
+            enforce_connectivity,
+            min_size_factor,
+            max_size_factor,
+            slic_zero,
+            start_label,
+            mask,
```
egpbos, line 155: Also here replace with `**kwargs` then (or `**slic_kwargs`).

egpbos (changes requested):
Awesome that you got this to work! I have a few comments to make the code a bit more compact and decrease maintenance burden.

dianna/methods/shap.py
geek-yang: Sounds good! Corrected!

dianna/methods/shap.py
geek-yang: Thanks for your comment @egpbos . It has been changed now.

dianna/methods/shap.py
geek-yang: Thanks for your comment @loostrum . Sounds good to me! I will move them to utils to make the code cleaner.

dianna/methods/shap.py
geek-yang: Thanks @egpbos for providing the solution! Indeed I was a bit confused about how to pick up the right argument for the `slic` function as we have already passed `kwargs` to the class when creating the object. I didn't notice that we've already got `dianna.utils.get_kwargs_applicable_to_function(slic, kwargs)`. I have removed those unnecessary arguments and now the code looks much much cleaner! Thanks agin. A big thumb up!!!

geek-yang:
> Nice work! I've added some comments to make it more consistent with the utils we use in the other methods. You're adding two models in the test_data folder, but I don't think those are actually used anywhere?

Thanks for your quick feedback @loostrum . Indeed only the onnx model is used for testing. The other pytorch model was some legacy file when testing deeplift for texts. I will remove it.

loostrum:
> Thanks for your quick feedback @loostrum . Indeed only the onnx model is used for testing. The other pytorch model was some legacy file when testing deeplift for texts. I will remove it.

I don't think the ONNX model is used either; only `mnist_model.onnx` is used, `mnist_model_tf.onnx` isn't.

dianna/methods/shap.py
geek-yang: Thanks for your comment @loostrum . This is actually very important. The function generates masks for the input image and it must satisfy the data type requirement of the input node of onnx model. It now uses the data type obtained from the onnx input node, which is in a way more generic.

geek-yang:
Tests on windows fail due to the `onnx-tf` package. Most likely it is due to the compatibility. Unless we can find any alternative to this, otherwise I would suggest we just ignore this for now.

egpbos:
We should then disable the test on Windows. Here you can find a way to do it: https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-test-functions

So, for instance, add to the top of `tests/test_kernelshap.py` (all of the Shap tests are failing, right?):
```py
if sys.platform.startswith("win"):
    pytest.skip("skipping tests on Windows due to onnx-tf incompatibility", allow_module_level=True)
```

dianna/methods/shap.py
geek-yang: Thanks for the tip and redirect me to the functions in lime and utils @loostrum. Now the KernelSHAP module also uses the xarray and the related functions in utils to prepare the input image data, similar to what you did in LIME.

dianna/methods/shap.py
geek-yang: Thanks for the comment. Just change it to `out[i] = image`, but didn't find a way to simplify the code due to the if statement nested in the loop and the mismatch of the shape. Any thoughts about that? Otherwise I would just leave it like this, though it is not perfect, for sure.

geek-yang:
> > Thanks for your quick feedback @loostrum . Indeed only the onnx model is used for testing. The other pytorch model was some legacy file when testing deeplift for texts. I will remove it.
> 
> I don't think the ONNX model is used either; only `mnist_model.onnx` is used, `mnist_model_tf.onnx` isn't.


`mnist_model.onnx` is used in the test (https://github.com/dianna-ai/dianna/blob/92-kernelshap-with-images/tests/test_kernelshap.py). `mnist_model_tf.onnx` is used in the tutorial (https://github.com/dianna-ai/dianna/blob/92-kernelshap-with-images/tutorials/kernelshap_mnist.ipynb) to showcase the 'capability' of kernelshap. Indeed I tried `mnist_model.onnx` for the tutorial, which is the one we trained using pytorch, but the result is not as obvious as `mnist_model_tf.onnx`, which is trained using tensorflow. I guess the reason is, `mnist_model.onnx` overfits a bit than `mnist_model_tf.onnx` and it influences the performance of kernelshap.

Currently we have many examples based on `mnist_model.onnx` (for instance, RISE). So I decide to keep both of them.

cwmeijer:
What's wrong with the Windows support? Skipping a test doesn't fix the actual functionality of course.
If we can't make it work for any of our supported platforms, we should discuss how to cope with this before merging.

cwmeijer (changes requested):
It should work smoothly with our API on all platforms. If not, let's discuss how to solve it.

(I will look at the files later, and provide you with some more helpful feedback if you like ;-P )

geek-yang:
> It should work smoothly with our API on all platforms. If not, let's discuss how to solve it.
> 
> (I will look at the files later, and provide you with some more helpful feedback if you like ;-P )

On windows the complaint comes from the onnx-tf package. It would be nice if you can test it on your windows machine and leave some remarks.

PR #109, opened 2022-01-17
Rename axes_labels parameter to axis_labels

loostrum:
Because axis should really be singular here and it's leading to confusion.


cwmeijer (approved):
haha, great PR! 👍
Merge it!

Issue #110, opened 2022-01-17
Cannot select label when using `dianna.explain_image`.


loostrum:
Unlike `dianna.explain_text`, `dianna.explain_image` does not have a `labels` kwarg.
Therefore one cannot choose a label when using the top-level interface, as any extra kwargs are 
fed to the method constructor and not to the method's `explain_image` method.
Perhaps we can use `utils.get_kwargs_applicable_to_function` to automatically feed the right 
kwargs to both the constructor and explain_x methods, so all kwargs can be used from the top-level interface
if one so desires.

loostrum:
from the standup: We will use the forward kwargs to function tool to fix this. label should be added as an explicit argument too, so we can keep documentation at the top-level for it.

loostrum:
Blocking image-based tutorials that use the top-level interface

loostrum:
Fixed by #114

PR #114, opened 2022-01-18
Allow for passing explain_image/text kwargs to top-level function

loostrum:
Kwargs to the top-level functions can now belong to either a method initializer or its explain_x.
Also added an explicit `labels` argument to explain_image. Not doing this with **kwargs, so we can keep the docstring for
`labels` in the top-level function.

To make this work, I made the following changes as well:
- allow for label selection in RISE.explain_image instead of always returning all labels
- allow for selection of more than one label in LIME.explain_image. LIME already produced results for multiple labels, but the get_image_and_mask function was run on only one label. Now it's run for all labels provided by the user

Now LIME and RISE return objects of the same shape.

### DIFF
--- a/dianna/__init__.py
+++ b/dianna/__init__.py
@@ -23,6 +23,7 @@ See https://github.com/dianna-ai/dianna
 """
 import logging
 from . import methods
+from . import utils
 
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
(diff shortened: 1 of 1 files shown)


loostrum:
Fixes #110

geek-yang (approved):
Looks pretty awesome and this makes the top level API more generic and adaptable. I will apply the same changes to `shap`, once the issue related to CI is solved. Just merge it.

PR #115, opened 2022-01-18
Tutorial notebook for RISE for images.

### LINKED ISSUE #98: Tutorial notebook for RISE for images.
1 example per notebook.
Use the dataset in the exploration repo. 
We want to cover binary MNIST. OR triangles and circles OR LeafSnap datasets and cover all in all notebooks.

geek-yang:
Two notebooks are added/revised to showcase the RISE implementation in DIANNA:
- Explain ImageNet with RISE
- Explain binary MNIST using RISE

These two notebooks are generated based on the old tutorial notebooks in DIANNA-exploration and DIANNA repo. Major changes include:
- Use high-level DIANNA API (instead of calling the method directly from `dianna.methods`)
- Add more explanation and remove unnecessary (testing/investigation) scripts
- Finetune model parameters and improve results

This PR also addresses issue #119 

Close #98


tutorials/rise_mnist.ipynb
```diff
@@ -5,14 +5,13 @@
    "id": "hispanic-instrumentation",
```
cwmeijer, line 5: <p>See my comment about 'fraction of the image' not being accurate in the other notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dianna-ai/dianna/pull/115/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cwmeijer:
<p>It seems I lost my comment in the imagenet notebook about the word 'fraction'.</p><p>It said that the phrase 'fraction of the image to keep' is only accurate on average, but not in general.</p><p>I think the phrase should be something like:</p><p>'... and for each mask and each feature in the image, the probability of being kept unmasked.'</p><p>It's hard to get that sentence right.</p><p> </p>

geek-yang:
> It seems I lost my comment in the imagenet notebook about the word 'fraction'.
> 
> It said that the phrase 'fraction of the image to keep' is only accurate on average, but not in general.
> 
> I think the phrase should be something like:
> 
> '... and for each mask and each feature in the image, the probability of being kept unmasked.'
> 
> It's hard to get that sentence right.

@cwmeijer Thanks for the comment. It is changed based on your suggestion.

tutorials/rise_mnist.ipynb
geek-yang: Corrected!

cwmeijer:
All notebook checks regarding rise are passing. @geek-yang and I agreed to merge it.

PR #111, opened 2022-01-17
Add LIME on images tutorial (fixes #97)

loostrum:
The output doesn't look right yet.
Also waiting for a decision on a few other issues that might be nicer to fix before the tutorials (see comments in #97)

PR #126, opened 2022-01-20
Tutorial notebook for KernelSHAP for images.

### LINKED ISSUE #95: Tutorial notebook for KernalSHAP for images.

geek-yang:
Two notebooks are added/revised to showcase the KernelSHAP implementation in DIANNA:
- Explain binary MNIST using KernelSHAP
- Explain geometric shapes with KernelSHAP

These two notebooks are generated based on the old tutorial notebooks in DIANNA-exploration and DIANNA repo. Major changes include:

- Use high-level DIANNA API (instead of calling the method directly from dianna.methods)
- Add more explanation and remove unnecessary (testing/investigation) scripts

Close #95


geek-yang:
The explanation of model used to classify geometric shapes is hard to interpret, even after quite some finetuning of parameters. This is an known issue during the exploration stage when we tried to explain it with `deeplifthshap` using the `captum` implementation. I would suggest we keep it for now and make another issue for later.

geek-yang:
Still cannot fix the import error in the test of notebook ` ImportError: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html`. Any idea about it?

loostrum:
I saw that error too, not sure what to do about it apart from trying the suggestions at the suggested URL. I haven't run into it on my laptop. Perhaps you can make an issue for it, if there isn't one already? 

> The explanation of model used to classify geometric shapes is hard to interpret, even after quite some finetuning of parameters. This is an known issue during the exploration stage when we tried to explain it with `deeplifthshap` using the `captum` implementation. I would suggest we keep it for now and make another issue for later.

I agree, let's make an issue and look into it later.

Is there a specific reason there are 2 notebooks for this method? I though we'd do one notebook for each modality/method combination. It would be a shame to throw away work put in the second notebook, so if we decide to keep one here, perhaps we could still store the other one in dianna-exploration.

loostrum:
Patrick fixed the ipywidgets issue in another PR (#111), so we should be good on that front.

egpbos:
Indeed, that one is fixed there. The fix was to just install ipywidgets :) So ece8305 can probably be reverted (if there's nothing else in there).

geek-yang:
> Indeed, that one is fixed there. The fix was to just install ipywidgets :) So [ece8305](https://github.com/dianna-ai/dianna/commit/ece8305907ebfd1b77518ce00c96f70dc5454f0e) can probably be reverted (if there's nothing else in there).
> 
> Edit: I see now you actually added ipywidgets to the dependencies as well :) All the other stuff mentioned in the link seems unnecessary though, so that can be taken out of the notebook again.

Thanks @egpbos for handling this :).

egpbos (approved):
LGTM! Tests pass on my laptop like this. Let's wait for CI and merge if everything is ok.

elboyran:
<ul><li>I propose to try another image with a triangle. That might end up being more interpretable (because it has sides and corners).</li><li>Wouldn't it be better to use the blue (not relevant) &lt;----&gt; red (relevant) colormap for our heatmaps? (In all notebooks.)</li></ul>

egpbos:
@elboyran do you think those things need to be changed in this PR or should we create a new issue for them and merge now?

elboyran:
For now 

> @elboyran do you think those things need to be changed in this PR or should we create a new issue for them and merge now?

For now we can merge, but there are conflicts to resolve first.

geek-yang:
> For now
> 
> > @elboyran do you think those things need to be changed in this PR or should we create a new issue for them and merge now?
> 
> For now we can merge, but there are conflicts to resolve first.

The conflicts are resolved. I will merge this branch. @elboyran I will add your comments into the existing issue #160 and get back to it later. Thanks for the suggestions!

PR #132, opened 2022-01-24
add notebooks extras_require for running notebooks

cwmeijer:
Currently the notebooks action in main is failing. One of the issues is that it misses the torchvision requirement. This PR fixes that issue. Other issues still remain after this PR, for instance, the missing ferrari.jpg (of which the license is unclear).
Feel free to merge if you approve. It's a tiny PR anyway.

### DIFF
--- a/.github/actions/install-python-and-package/action.yml
+++ b/.github/actions/install-python-and-package/action.yml
@@ -23,4 +23,4 @@ runs:
       shell: bash {0}
       run: |
         python3 -m pip install --upgrade pip setuptools
-        python3 -m pip install .[dev,publishing]
+        python3 -m pip install .[dev,publishing,notebooks]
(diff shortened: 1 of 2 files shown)


egpbos:
Hm, shouldn't the notebook action install dependencies from `requirements.txt` instead? Sorry for the sloppiness on my part in the original notebook action PR!

cwmeijer:
We don't have such a file at the moment. Are you proposing to switch methods regarding dependency listings in this PR?

egpbos:
Ah, I was confused with dianna-exploration, I was thinking of this PR: https://github.com/dianna-ai/dianna-exploration/pull/145

I think it's useful to do this specifically for the notebooks, because then they also work on Binder. Binder allows users to try out the tutorial notebooks in the cloud for free.

Anyway, maybe just merge this and we can change it later if we want.

PR #157, opened 2022-01-25
add tutorial notebooks to documentation (refs #100)

### LINKED ISSUE #100: Investigate showing tutorial notebooks in documentations.
Ask Stef as well.

cwmeijer:
Fixes #100.


egpbos:
Fixes #100.

docs/tutorials
```diff
@@ -0,0 +1 @@
+C:/Users/ChristiaanMeijer/Documents/dianna/dianna/tutorials
```
loostrum, line 1: I'm pretty sure this won't work apart from on Christiaan's machine ;) 
We could use nblink instead?

cwmeijer: haha, I can see why you'd be suspicious about that line, but have you tried running it? ;-P

Anyway, we hoped to fix this with a symlink (like we tried here) but my experience with symlinks on Linux is already limited but on Windows even less. I guess this was my first attempt to make one on Windows.
We choose not to use nbsphinx_link because that would mean we'd need to do bookkeeping of our notebooks twice, which is errorprone. Also, because the package was not maintained the last 2 years, we preferred to not rely on it.

I think I'll try to fix the symlink first. If that doesn't work, we can switch to nbsphinx_link anyway.

cwmeijer: Apparently making a relative link to a directory is some cutting edge feature that my current Powershell cannot handle yet. I'll check if I can get my hands on some newer version later (tomorrow probably) to test that out. I have PS5 and apparently I'd need PS7.1 (https://stackoverflow.com/questions/42513791/how-to-create-relative-symlink-powershell-way)

egpbos: Meine Güte. I'm reading up on teh Googlez a bit and it seems symlinks in git on Windows in general is all still iffy and annoying (needs special rights, maybe git settings, git hooks...). I'm starting to wonder whether it's worth the effort. Maybe revert to using `nbsphinx_link` after all?

loostrum: Unfortunately the windows-type relative symlink doesn't work on Mac/Linux. We only needs this to build the documentation anyway, right? We could limit the doc build action to only one OS and use that to deploy it. But probably going to nbsphinx_link is easiest...

egpbos: That is a good idea as well! But I think I agree that using nbsphinx_link is slightly better, since that way Windows devs can also still build docs locally.

cwmeijer: I managed to create a symlink in PS7.1. It clearly shows it's there. But even then, I haven't even figured out how to use it, it keeps showing up in listings as a link, without showing what's behind it. It's a pain really. Shame I only read your comments just now, otherwise I probably would have stopped trying an hour ago.

Anyway, I'm now going for the nbsphinx_link approach again. Thanks for your thoughts and suggestions.

egpbos:
Ah, we have to install pandoc separately, the Python package only provides a Python interface to the actual pandoc program/library.

Perhaps we should make a separate docs workflow which only runs on Ubuntu to make this easier (we can just do `apt install pandoc`). What do you think?

cwmeijer:
Ah right, now I remember. I had to install pandoc through an msi installer here locally to be able to run the workflow. Then it makes sense that it won't run on the server.

Your solution sounds great! I'll try that on Monday.

loostrum:
On windows on the CI we should be able to install pandoc with `choco install pandoc`, but perhaps it's easier to stick to one OS for the docs

cwmeijer:
@egpbos or @loostrum, it now passes all checks. If either one of you can have a quick glance at my very first github action (I think), then it should be ready to be merged.
I'm only building the documentation on a single os and python version in order not to clutter our checklist even more.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,31 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest']
+        python-version: ['3.9']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/install-python-and-package
+        with:
+          python-version: ${{ matrix.python-version }}
```
egpbos, line 26:
```suggestion
    name: Build documentation
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
    steps:
      - uses: actions/checkout@v2
      - uses: ./.github/actions/install-python-and-package
```

Slight simplifications: 1) no need for a matrix anymore and 2) no need to specify Python version (the default argument is already 3.9).

setup.cfg
```diff
@@ -66,6 +66,9 @@ dev =
     torchtext
     coverage [toml]
     nbmake
+    nbsphinx
+    nbsphinx_link
+    pandoc
```
egpbos, line 71: Do we still need the `pandoc` dependency here?

egpbos (commented):
Looks good!

PR #159, opened 2022-01-25
Cleanup root directory

### LINKED ISSUE #120: Cleanup files
Now that we're making everything nice and neat, there is one thing that pains my eyes greatly and that is the huge-a-mongous list of files in our top level repo directory which hides the readme on GitHub 40 wheel-scrolls down the page, making all our painstaking work on the readme useless. I think we should do a radical cleanup of the top-level dir. Remove all unused files and put everything that we can in other directories.

egpbos:
Remove unused files left over from the Cookiecutter template and move a movable configuration file. Fixes #120.


loostrum (changes requested):
I love it, the root dir looks much better now.

I think we do need .zenodo.json though, it's used by Zenodo to automatically generate a release when we make one on Github (and it needs to be in the root too :/). The cffconvert action should complain, but it seems broken as it just prints the usage of the `cffconvert` command

egpbos:
I thought .zenodo.json was made obsolete by Zenodo's CFF support, see e.g. https://twitter.com/zenodo_org/status/1420357001490706442.

loostrum (approved):
Ah nice, I didn't know about that. Then the PR is good to go!

PR #167, opened 2022-01-27
optimize CI pip dependency installation

egpbos:
This commit/PR:

1. Adds the option in the install action to specify which extras-require options to use (dev, notebooks, publish).
2. Uses the above to minimize the dependencies that have to be installed for each workflow.
3. Installs wheel in the first step of the dependencies installation. This allows pip to download wheels in the second step (installing dianna and dependencies), which reduces install time.


loostrum (approved):
Looks good, faster CI, fewer dependencies: I like it. Merge it!

PR #168, opened 2022-01-27
Developer documentation

### LINKED ISSUE #146: readme/docs: add development section
Following the GT suggestions in #116, we should add a development section to the readme with the following contents:

- Details about development environment and test environment
- Up to date developer documentation (see README.dev.md)

I think we should keep extra files to a minimum, so I would suggest the README.dev.md file (if we still have it) should be part of the full documentation. We can just make a page/chapter there about all the development details.

egpbos:
Fixes #146 

This puts developer information in the readthedocs.


egpbos:
Note that the link checker CI will fail, because the readthedocs page does not yet exist. Hopefully I guessed the URL correctly 😄

docs/developer_info.rst
```diff
@@ -0,0 +1,213 @@
...
+   source env/bin/activate
+
+   # make sure to have a recent version of pip and setuptools
+   python3 -m pip install --upgrade pip setuptools
+
+   # (from the project root directory)
+   # install dianna as an editable package
+   python3 -m pip install --no-cache-dir --editable .
+   # install development dependencies
+   python3 -m pip install --no-cache-dir --editable .[dev]
```
loostrum, line 28: Only the `.[dev]` install is needed, right? Or do you also want to show how to install without dev dependencies?

docs/developer_info.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Afterwards check that the install directory is present in the ``PATH`` environment variable.
```
loostrum, line 30: I think you mean `PYTHONPATH`, but even then I don't think it's actually in there. In my experience at least Python creates a link in the `site-package` folder to wherever the original code is. I don't know if it always works like that though

docs/developer_info.rst
```diff
@@ -0,0 +1,213 @@
...
+Dependencies and Package management
+-----------------------------------
+
+DIANNA aims to support all Python 3 minor versions that are still
+actively maintained, currently:
+
+-  3.7
+-  3.8
+-  3.9
+-  3.10 is still missing pending the availibility of onnx-tf in 3.10.
```
loostrum, line 42: onnxruntime I think actually? Or perhaps onnx-tf as well

docs/developer_info.rst
egpbos: Right, will edit.

docs/developer_info.rst
egpbos: Good catch, I'll just take the whole sentence out, seems useless with modern `venv` and `conda`.

docs/developer_info.rst
egpbos: Right!

loostrum (approved):
Awesome, merge time!

loostrum:
I had to re-run the workflow because the markdown link check was faster than the readthedocs build, but now everything is green so you guessed the URL right :)

Issue #99, opened 2022-01-11
Tutorial notebook for RISE for text.


elboyran:

loostrum:
Closed #99 by #173

PR #173, opened 2022-01-28
Add RISE on text tutorial

loostrum:
The results are not amazing, but it's working.
I think we might want to change something in how RISE masks sentences, but I'll open a separate issue for that.


egpbos:
This fixes #99, right?

egpbos (approved):
The notebook looks good in general. Only comment I would have is to slightly reword the results section which mentions "words against the selected class are blue", while no such words are present. This is also explained below the image, but it still tripped me up, so just a minor thing to consider perhaps.

loostrum:
Thanks, I'll reword that and then merge

PR #177, opened 2022-01-28
Readme: statement of need

elboyran:
Addresses #144 and #138 .

PR #178, opened 2022-01-31
added high level overview to readme

### LINKED ISSUE #137: readme(/JOSS): add high-level summary
For JOSS we need: "A summary describing the high-level functionality and purpose of the software for a diverse, non-specialist audience."

Related, the GT suggested these general comments for the readme which may be relevant:

> The README intro reads too much like a paper intro. It should be shorter and more to the point, then extended with the current information. It is not clear from the README explanation who is target user of the package, and what software stack it supports. In the first paragraph, it could have this information. For instance, “DIANNA provides insight tools for Deep Learning frameworks in order to increase the explainability of these models. It supports PyTorch and Tensorflow in addition to blah.”
> 
> Furthermore, you need to be more explicit about “non-specialist researcher user”. What experience does the user need to use DIANNA? For example experience with PyTorch/Tensorflow, ONNX and familiarity with Explainable AI? You should define the target user profile.
> 
> Anything that is “new” to the user could be reinforced. This could also help promote the DIANNA docs as the first contact about some of these topics, e.g. “Have you heard about XAI? See how to make your Deep Learning model explainable here.”

egpbos:
Fixes #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,8 @@ affiliations:
 
 # DIANNA: Deep Insight And Neural Network Analysis
 
-DIANNA is a Python package that brings explainable AI (XAI) to your research project.
-
-- Wraps carefully selected XAI methods in a simple, uniform interface.
-- Uses [ONNX](https://onnx.ai/) model format: interoperable with all the big ML frameworks (e.g. PyTorch, TensorFlow).
-- Comes with a benchmark set of models and datasets: compare and gain quick insight into different XAI methods.
-- Built by, with and for (academic) researchers and research software engineers working on machine learning projects.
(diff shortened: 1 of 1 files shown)


elboyran:
Some checks are not passing, is it ready for review?

loostrum (approved):
I like it! Short and to the point, which is great for potential users

elboyran (changes requested):
As you noticed already, it now overlaps with what's under Why DIANNA below (after merging #177). I propose we leave one sentence (merging first and last bullet points as I didn't mention them in #177) and add informative image possibly with 1 sentence.

elboyran (approved):

Issue #174, opened 2022-01-28
Change method of RISE text masking


loostrum:
Currently, as masked word is simply removed from the input sentence.
LIME does this differently: it replace a masked word by "UNKWORDZ", which a typical tokenizer will then throw out and replace by a special "unknown word" token.

Both are valid methods for masking, but the distinction is important when e.g. a CNN is used that depends on neighbouring words, like the movie review model that we have. To be consistent, I suggest that we change our RISE implementation to also replace masked words by a special (user-definable) string.

PR #185, opened 2022-02-01
In RISE for text, mask tokens with a user-definable string instead of removing  them

loostrum:
The default mask string is "UNKWORDZ", which most likely isn't in any tokenizer's dictionary.
This is the same value as is used internally in LIME. The name of the parameter (`mask_string`) is also identical to LIME, the DIANNA API remains consistent with respect to this parameter.

### DIFF
--- a/dianna/methods/rise.py
+++ b/dianna/methods/rise.py
@@ -19,7 +19,7 @@ class RISE:
     required_labels = ('batch', 'channels')
 
     def __init__(self, n_masks=1000, feature_res=8, p_keep=None,  # pylint: disable=too-many-arguments
-                 axis_labels=None, preprocess_function=None):
+                 axis_labels=None, preprocess_function=None, mask_string="UNKWORDZ"):
         """RISE initializer.
 
         Args:
@@ -30,6 +30,7 @@ class RISE:
                                                If a list, the name of each axis where the index
                                                in the list is the axis index
             preprocess_function (callable, optional): Function to preprocess input data with
+            mask_string (str, optional): String to replace masked tokens with (text only)
         """
         self.n_masks = n_masks
         self.feature_res = feature_res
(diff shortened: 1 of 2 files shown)


loostrum:
This PR closes #174

loostrum:
The markdown link checker complains about nlp.standford.edu, which does seem to be down at the moment. I'm assuming that's temporary, and it's unrelated to this PR anyway.

dianna/methods/rise.py
```diff
@@ -115,7 +117,7 @@ def _get_predictions(self, sentences, runner, batch_size):
     def _create_masked_sentences(self, tokens, masks):
         tokens_masked = []
         for mask in masks:
-            tokens_masked.append(tokens[mask])
+            tokens_masked.append([token if keep else self.mask_string for token, keep in zip(tokens, mask)])
```
cwmeijer, line 120: I love this line!
I have to admit though, that it took a couple of minutes before it snapped into place in my mind. So on one hand it looks beautiful, but it isn't the clearest of lines. I could not think of a better alternative, other than extracting the list comprehension to its own function. This wouldn't solve much I think except for reducing the complexity on a single line. I am not sure if we should change it as it is.

cwmeijer (approved):
Great that you took so much care to keep the API consistent over all XAI methods.
Excellent addition! 🥇

tests/test_rise.py
```diff
@@ -57,7 +57,7 @@ def test_rise_text(self):
         review = 'such a bad movie'
         expected_words = ['such', 'a', 'bad', 'movie']
         expected_word_indices = [0, 5, 7, 11]
-        expected_positive_scores = [0.3295266, 0.3521292, 0.023648001, 0.3347813]
+        expected_positive_scores = [0.3044678, 0.28736606, 0.03623142, 0.23650846]
```
cwmeijer, line 60: Seems like the general trend in the values has remained unchanged. That's great!

dianna/methods/rise.py
loostrum: Thanks! I agree it's not the most readable but like you I couldn't think of a better way yet. I'll leave it as-is for now.

PR #186, opened 2022-02-01
Remove need for batch axis in input images

### LINKED ISSUE #88: Remove need for batch axis in input
Currently we require that the user provides data with a batch axis.
However, the batch size is always one as only one image/text can be explained at a time.
We should require data _without_ batch axis instead, and add it internally where need (e.g. for RISE)

loostrum:
Until now, image data needed a batch axis in the user input. This is now removed. Where needed, a batch axis is added internally in DIANNA. Now the API matches that of text-based methods, which also didn't require a batch axis.

The image-based tutorial notebooks were updated to remove the batch axis as well.

This PR closes #88


cwmeijer (approved):
All looks great. Great API improvement. Notebooks and tests look much clearer. 🎖🥇🥈🥉🏅

PR #187, opened 2022-02-01
Add preprocess function option to kernelshap

loostrum:
See #155

### DIFF
--- a/dianna/methods/kernelshap.py
+++ b/dianna/methods/kernelshap.py
@@ -149,6 +149,8 @@ class KernelSHAP:
         Returns:
             transformed input data
         """
+        if self.preprocess_function is not None:
+            input_data = self.preprocess_function(input_data)
         input_data = utils.to_xarray(
             input_data, self.axes_labels, KernelSHAP.required_labels)
         # remove batch axis from input data; this is only here for a consistent API
@@ -213,6 +215,5 @@ class KernelSHAP:
                                        self.channels_axis_index,
                                        self.input_node_dtype.as_numpy_dtype
                                        )
-        if self.preprocess_function is not None:
-            model_input = self.preprocess_function(model_input)
+
         return prepare(self.onnx_model).run(model_input)[f"{self.output_node}"]
--- a/dianna/methods/kernelshap.py
+++ b/dianna/methods/kernelshap.py
@@ -11,14 +11,16 @@ class KernelSHAP:
     # axis labels required to be present in input image data
     required_labels = ('batch', 'channels')
 
-    def __init__(self, axis_labels=None):
+    def __init__(self, axis_labels=None, preprocess_function=None):
         """Kernelshap initializer.
 
         Arguments:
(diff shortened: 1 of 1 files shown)


dianna/methods/kernelshap.py
```diff
@@ -204,13 +206,13 @@ def _runner(self, features):
...
-            )
-        )[f"{self.output_node}"]
+        model_input = self._mask_image(features,
+                                       self.image_segments,
+                                       self.input_data,
+                                       self.background,
+                                       self.channels_axis_index,
+                                       self.input_node_dtype.as_numpy_dtype
+                                       )
+        if self.preprocess_function is not None:
```
geek-yang, line 216: I wonder if it makes more sense to move the `preprocess_function` before calling the segmentation (line 122). If the user wants to preprocess the input data, I guess it happens before the segmentation and masking. What do you think @loostrum ?

geek-yang (approved):
Thanks for implementing the `preprocess_function`. Just a small comment about the location of this preprocess procedure.

dianna/methods/kernelshap.py
```diff
@@ -147,6 +149,8 @@ def _prepare_image_data(self, input_data):
         Returns:
             transformed input data
         """
+        if self.preprocess_function is not None:
```
geek-yang, line 152: I moved the `preprocess_function` to the function `_prepare_image_data`. There the `preprocess_function` is executed before the processing of input data inside `kernelshap` and therefore the user does not need to be bothered by the operations inside `kernelshap` when implementing their preferred way of preprocessing the input image. I think this is more friendly to the user.

geek-yang (approved):

geek-yang:
And BTW, thanks for keeping the API consistent for all methods. I really love this kind of changes you made to our repo @loostrum!

dianna/methods/kernelshap.py
loostrum: I'm not sure I get why you moved it all the way to the start of the processing. What would then be the point of having a preprocessing function? The user could then just do it themselves before using dianna. 
In the other methods, the preprocess function is called as late as possible, just before running the model. The reason being that the model might require slightly different input than the XAI method. E.g. Images must have a channels axis, which you might not have for greyscale data. Then you could add the channels axis, and add a preprocessing function that removes the channels axis. Such a preprocessing  function should only be called just before running the model. 
You're absolutely right that it requires knowledge of the internals of the method, but I think that is exactly the point in this case

geek-yang: Hi @loostrum , thanks for your explanation. My impression about `preprocess` comes from some other python packages (for instance, like [xarray.open_mfdataset](https://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html). The process begins before the operations regarding the input. But I totally agree with you in this case and I think what you suggest here make more sense. Thanks again for the clarification and sorry for making this quite confusing. I will revert my change and just go ahead and merge it.

loostrum: No problem, thanks for picking up this PR again!

Issue #127, opened 2022-01-20
Conversion to ONNX 


elboyran:
Take the ONNX conversion notebooks from dianna-exploration

PR #194, opened 2022-02-03
add more information to onnx section and pointers to conversion notebooks

geek-yang:


elboyran (changes requested):
@geek-yang Did you plan to also put links for the notebooks where we create our models? At dianna-exploration?
Because now I'm missing those. And also not sure the sentence above the models for benchmark data really corresponds to what's below, namely the conversion scripts, not the model generation (the missing info).

Or maybe we need small table with Zenodo links to the models and generation scripts for each dataset?

I have also just had the idea that maybe we can point also to the ONNX tutorial page as there they cover even more converters with examples and tutorials (I should have checked that before to save us some hours...): https://github.com/onnx/tutorials

geek-yang:
> @geek-yang Did you plan to also put links for the notebooks where we create our models? At dianna-exploration? Because now I'm missing those. And also not sure the sentence above the models for benchmark data really corresponds to what's below, namely the conversion scripts, not the model generation (the missing info).
> 

@elboyran Thanks for the comment! I thought this issue only addresses the onnx conversion notebook (was misled a bit by the title and its blocking tag due to #127). And yes, about the mismatch of the sentence above the models for benchmark data and the links below, sorry for the confusion.

> Or maybe we need small table with Zenodo links to the models and generation scripts for each dataset?
> 

That's a good idea! I will make a table with Zenodo links to the models, generation scripts for each dataset, and the conversion notebook.

> I have also just had the idea that maybe we can point also to the ONNX tutorial page as there they cover even more converters with examples and tutorials (I should have checked that before to save us some hours...): https://github.com/onnx/tutorials

Thanks for mentioning this! I think this is good for the user (not anybody's fault for noticing this a bit late, only because those onnx people don't sell this well 😉 :wink:). I will point to these examples.

elboyran (approved):
Good work!

PR #198, opened 2022-02-04
Explicitly set setup-python version

loostrum:
Note: the version should be made less specific again once the general v2 tag is updated to v2.3.2 or later of setup-python


loostrum:
Already fixed upstream, v2 now points to v2.3.2. See https://github.com/actions/setup-python/issues/328

PR #211, opened 2022-02-08
improve results in tutorial notebook kernelshap with geometric shapes

geek-yang:


dianna/methods/kernelshap.py
```diff
@@ -175,7 +173,7 @@ def _mask_image(
         """
         # check the background color
         if background is None:
-            background = image.mean((0, 1))
+            background = image.mean(axis=(0, 1))
```
geek-yang, line 176: By default, the mean of an xarray object will take the first argument as labels, but here we need axis. So is the change.

elboyran (changes requested):
Very nice, happy to see the triangle worked better! ;-)

If it's not much work (I think it's just changing the colormaps), could you, please, change the input images to gray-scale and the relevance scores to be in blue (not relvant) <---> red (relevant). See the informative image in our README now.

geek-yang:
> Very nice, happy to see the triangle worked better! ;-)
> 
> If it's not much work (I think it's just changing the colormaps), could you, please, change the input images to gray-scale and the relevance scores to be in blue (not relvant) <---> red (relevant). See the informative image in our README now.

@elboyran Thanks for your comment! Now the input images in kernelshap notebooks are shown in gray scale and the color map for explanation is blue-red.

elboyran (approved):
Nice!

Issue #206, opened 2022-02-08
Improve explanation results in the tutorial notebooks of RISE text


gcroci2:
Results from rise_text.ipynb:
<img width="1137" alt="Screenshot 2022-02-08 at 12 07 05" src="https://user-images.githubusercontent.com/55382553/152975652-74f9f15b-0704-4735-a934-87a2cc8c37a9.png">

Results from lime_text.ipynb:
<img width="1129" alt="Screenshot 2022-02-08 at 12 07 16" src="https://user-images.githubusercontent.com/55382553/152975662-b9a83e09-13aa-4119-be48-9059ffa0c0d3.png">

If you look at the values, results are actually similar in terms of scores, but the scores range is different. We need to figure out scores range and how this can be solved.

elboyran:
@gcroci2 I do not get what you mean - I see different scores for the same words in both methods. It's been a while, but if you wish, could you,please, elaborate?

gcroci2:
> @gcroci2 I do not get what you mean - I see different scores for the same words in both methods. It's been a while, but if you wish, could you,please, elaborate?

Since they are two different methods, the scores assigned in each of them may have a different range. When we plot the highlighted text, we do not normalize the two different ranges, so they are not really comparable, and also the plots can be more confusing. If for example Rise has a range between 0 and 1, and Lime has a range between 0 and 0.5, before plotting them we should use some normalization function to match the ranges. This is of course not needed if we're sure that both methods are giving scores using the same range (or domain).

PR #214, opened 2022-02-09
add informative screenshots in the readme file

gcroci2:


elboyran (changes requested):
Somewhere in the image captions or axis labels should be clear that:

- this is from Binary MNIST, not MNIST
- which class is this explanation for
- I've asked many times, but still SHAP uses a different colormap :-(. I would like blue- negative relevance- red- positive for all methods...

elboyran (commented):
In the related issue I see also a leaf image and relevances, is the idea to add it too? Or instead? 

I wonder do we also want actually 1 image and 1 sentence ( to claim data modality generality to some modest extent ;-)

elboyran:
> Would that work?
> 
> dianna_mnist
> 
> For the leaf image, I think it would be nice to add it! But I am not sure about the impacts of the results because not all of them makes a clear sense for the leaf model explaination (standup tag?)
> 
> Regarding the sentence yes, I think it'a good idea to add it too. But also in this case, not all the methods gave satisfactory results (issue #206)

To me this looks nicer than before, thanks! ;-). I don't want to sound pedantic (and sorry for not spotting it earlier), but MNIST (and it's so famous, everybody knows it ;-) ) originals are gray scale (look like white digits on black background), ours now is in green on purple background, could you, please, change it to white on black? I think it should be just a mapper of gray-scale figure colormap.

gcroci2:
> > Would that work?
> > dianna_mnist
> > For the leaf image, I think it would be nice to add it! But I am not sure about the impacts of the results because not all of them makes a clear sense for the leaf model explaination (standup tag?)
> > Regarding the sentence yes, I think it'a good idea to add it too. But also in this case, not all the methods gave satisfactory results (issue #206)
> 
> To me this looks nicer than before, thanks! ;-). I don't want to sound pedantic (and sorry for not spotting it earlier), but MNIST (and it's so famous, everybody knows it ;-) ) originals are gray scale (look like white digits on black background), ours now is in green on purple background, could you, please, change it to white on black? I think it should be just a mapper of gray-scale figure colormap.

mnist_comparison

Please suggest other modifications if needed!

elboyran:
Nice!

PR #215, opened 2022-02-09
Automatically add channels for LIME if the input is greyscale

loostrum:
Even if the input is greyscale, like MNIST, LIME still requires 3 channels. The preprocessing function was our workaround to handle this, but it is unclear to the user because LIME adds a batch axis, before the preprocessing function is called.

This PR adds functionality to automatically add/remove the extra channels for LIME.

I'm not too happy about the `_get_full_preprocess_function` method, but I couldn't find a better way yet without assigning lambda expressions, which the linter doesn't allow.


egpbos (approved):
No comments, I like it much better this way! No more workarounds in the test case is always a good sign of improved UI.

Issue #149, opened 2022-01-24
more basic tutorial notebooks?


egpbos:
The GT suggested in #116 that we have more basic tutorial notebooks:

> [...] just give good titles, e.g., “Introduction to DIANNA” can be a tutorial, with an overview of everything; “Explainable AI using Tensorflow and DIANNA” can be a tutorial, more focused. “How to choose which DIANNA method to use” can go more to the point.

I'm not sure about how this would work. It seems to make sense from a user perspective, though, so let's discuss.

elboyran:
At least 1 basic and maybe 1 advanced generic tutorial notebooks.

elboyran:
Blocking issue #196

elboyran:
First the PR #223 should be resolved!

PR #223, opened 2022-02-10
add getting started to readme

cwmeijer:
We solved #149 by creating a 'getting started' section in the readme instead of another tutorial. We also moved the 'how to use dianna' section to the docs because it overlapped with the new section.


elboyran (changes requested):
Nice intro! (Unfortunately not enough for also submitting to ONNX repo, I guess we should address this at the next sprint.)

* I still think the schematic you removed should be in the front page README. "An image speaks 1000 words."
* I would still add a short encouragement to see the ONNX section about conversion from common formats. We don't want to scare people away from the very fist Getting started section.
* For the images example, I would like to have one of our datasets (as well?)
* The "generic" text and image examples is a nice idea, but then the explanations are not intuitive. Why some word in the Latin sentence are blue, other's - in red? 

1.  One way could be to specify the class we would like to have the explanation for- e.g. triangle for a triangle image. "Positive" for a movie review. 
2. Another way would be to further explain the graphic explanations shown in order to make some intuitive sense. By 1 sentence below or so.

* Long running comment within a cell is not very convenient for understanding - have to scroll. put it above the line, it should also help with the explanation above.

elboyran (approved):
It's better now, I think.

PR #240, opened 2022-02-17
revisit joss skeleton

gcroci2:
I removed Mathematics and Citations sections (I left References at the end as is done in the [recent paper example](https://joss.theoj.org/papers/10.21105/joss.03951)), and as it seems to be suggested in the [review criteria](https://joss.readthedocs.io/en/latest/review_criteria.html). I also inserted State of the field and Key Features sections. 

I left Figure section, also if I don't think is needed (see [recent paper example](https://joss.theoj.org/papers/10.21105/joss.03951))

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -21,7 +21,6 @@ affiliations:
    index: 3
 date: dd Month yyyy
 bibliography: paper.bib
-
 ---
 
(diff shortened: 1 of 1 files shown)


elboyran (approved):
It's fine, thanks.

PR #267, opened 2022-03-15
Joss paper

elboyran:


elboyran:
fair-software checks fail?

elboyran:
See PR #269 for possible fixing of the fair-workflow. After it also passes.

PR #269, opened 2022-03-17
Update fair-software.yml

jspaaks:
Patched the howfairis github action, then updated the workflow here. Seems to work now:

https://github.com/dianna-ai/dianna/runs/5592260896?check_suite_focus=true


elboyran:
@jspaaks, but many other checks now fail :-(

cwmeijer:
```python
     import tensorflow as tf
```

elboyran:
Merging in order to have the fair-software fix for the paper.

Issue #254, opened 2022-03-04
Make LIME work (again) with binary mnist


gcroci2:
Data I used are `binary-mnist.npz`, and the model is `mnist_model_tf.onnx`. See the [demo.ipynb](https://github.com/dianna-ai/dianna/blob/main/tutorials/demo.ipynb) file. 

<img width="1069" alt="image" src="https://user-images.githubusercontent.com/55382553/156757238-504f8413-37fc-4faf-9f8a-c960ccfde8cb.png">

LIME seems to be "broken" and it outputs all ones for this binary mnist image. BUT when I did this weeks ago, I obtained this nice one for the readme:

<img width="182" alt="image" src="https://user-images.githubusercontent.com/55382553/156757551-14168200-3fa1-4c3b-a1d6-d1578f9b3f84.png">

It's not clear if it's LIME itself or @loostrum fix to the grayscale images, but I also tried to add the channels and reproduce what I did the first time, before his fix, and it didn't worked as well.

egpbos:
After having confirmed that LIME still works (see #208), maybe it could help to take a look here: https://github.com/marcotcr/lime/blob/master/doc/notebooks/Tutorial%20-%20MNIST%20and%20RF.ipynb

egpbos:
The PR from @loostrum you mean is #215, right?

egpbos:
Tried out reverting to the repo's state from before #215 in #271: didn't work, still only 1's. Seems like it wasn't our doing then (or at least not our doing in that PR).

egpbos:
For reference, this is the commit where apparently things were still working: https://github.com/dianna-ai/dianna/commit/66ed07ca0d067c0286ae4ca8851b99c58418dba9. It was added in #214, so indeed one before #215. @gcroci2, were you using the master version at that time, or some slightly earlier version? In the latter case, we can still try reverting some other changes between when you used it and #215.

egpbos:
#272 reverts the repo to the state of Feb 1st. The demo notebook still only gives ones for LIME. @gcroci2, were you already working with DIANNA before Feb 1st? If not, the this should exclude the possibility that it was something we changed in DIANNA that broke LIME.

gcroci2:
[...]

PR #271, opened 2022-03-21
[DO NOT MERGE] Fixing #254: revert to before #215 (added grayscale only handling)

egpbos:
Added demo back in and tried it out in this version.

Result: LIME still gives only ones.

### DIFF
--- a/tutorials/demo.ipynb
+++ b/tutorials/demo.ipynb
@@ -1,0 +1,644 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


gcroci2:
Should we just leave it as it is (so not merging with the main branch) since it does not solve the problem? Same question for PR #272

egpbos:
Yes, indeed, I should have added a DO NOT MERGE tag, done now.

PR #273, opened 2022-03-22
Update setup.cfg

### LINKED ISSUE #270: tests failing: ModuleNotFoundError: No module named 'tensorflow_probability'

```python
import tensorflow as tf
```

cwmeijer:
fixes #270


cwmeijer:
A better fix was implemented in PR #274 and was already merged.

PR #274, opened 2022-03-22
add dependency (fixes #270)

### LINKED ISSUE #270: tests failing: ModuleNotFoundError: No module named 'tensorflow_probability'

cwmeijer:


elboyran:
@cwmeijer is it still work in progress?

PR #275, opened 2022-03-22
add youtube movie demo to readme

egpbos:


elboyran (approved):
Good addition. Hopefully the other PR fixes the failed builds.

cwmeijer:
Alternatively, we could put the video on the very top of the readme, right above the screenshots. I think that would be quite cool. The current video was of course not really made with such a purpose in mind. Maybe a 1 minute video with fewer details would be more fitting for putting on top of the readme. We could consider making a new one in a few months when we have some new features.

I think the PR is a great addition as it is.

cwmeijer (approved):
Great!, see my comment 1 second ago.

PR #276, opened 2022-03-22
LIME in demo fixed by using RGB values (0-255)

### LINKED ISSUE #254: Make LIME work (again) with binary mnist

egpbos:
Fixes #254.

The problem was that the input data was divided by 255, which maps the data to the [0,1] range, which is necessary for the models, but LIME expects "RGB values", i.e. in the [0,255] range.

The demo now uses a `preprocess_function` for LIME to fix this.


elboyran (approved):
Lets merge it and release.

PR #289, opened 2022-03-29
Update CI

egpbos:
Three updates in this PR:

1. Update GitHub actions versions (necessary because the old ones will soon stop working).
2. Simplify the release workflow: use the `build` module to build wheels and source distribution in one go, as recommended here: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives
3. Streamline the release workflow: only trigger the test.pypi upload when manually running the workflow. Before, it would also trigger on releases, but this is unnecessary. This way, the test.pypi.org channel can also be used to test an upload without having to bump the version.

### DIFF
--- a/.github/actions/install-python-and-package/action.yml
+++ b/.github/actions/install-python-and-package/action.yml
@@ -13,7 +13,7 @@ runs:
   using: "composite"
   steps:
     - name: Set up Python ${{ inputs.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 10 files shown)


egpbos:
Note that CI is failing due to this upstream error: https://github.com/tiangolo/typer/issues/377, which will be fixed by this PR: https://github.com/tiangolo/typer/pull/375 so we should wait for that.

tiangolo:
Hello! I just released Typer `0.4.1` that should handle this. :rocket: :nerd_face:

egpbos:
Awesome, thanks for the ping @tiangolo!

PR #316, opened 2022-06-29
use load_img from keras.utils to fix compatibility with tensorflow 2.9

### LINKED ISSUE #314: Tutorial notebooks no longer working on CI
The `rise_imagenet.ipynb` notebook fails with `AttributeError: module 'keras.preprocessing.image' has no attribute 'load_img'`

`https://stackoverflow.com/questions/72383347/how-to-fix-it-attributeerror-module-keras-preprocessing-image-has-no-attribu` suggests it may be due to tensorflow 2.9. We can limit ourselves to 2.8, or use a different function to load the image, `tf.keras.utils.load_img `

loostrum:
Fixes #314


loostrum:
Apparently there has been some update to the linter, because it is giving warnings about the dashboard (also on main, which still passed on 5 May).
In this PR I've fixed those warnings, and a whole bunch of warnings that only showed up locally.

PR #324, opened 2022-09-01
expose SimpleModelRunner to user

cwmeijer:
I think a user is not able to import the SimpleModelRunner at the moment. That must be a mistake. This commit exposes it to the user so they can reuse it.

I think this PR can be approved if any reviewer agrees on both questions below:
Do you agree that the class could not have been imported by a users before adding it to the __init__?

Do you agree that we want to make this part reusable?

### DIFF
--- a/dianna/utils/__init__.py
+++ b/dianna/utils/__init__.py
@@ -4,5 +4,5 @@ from .misc import get_kwargs_applicable_to_function
 from .misc import move_axis
 from .misc import onnx_model_node_loader
 from .misc import to_xarray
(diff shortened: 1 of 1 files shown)


loostrum (approved):
I think this mostly changes how SimpleModelRunner could be imported.
Before this PR, it would be
`from dianna.utils.onnx_runner import SimpleModelRunner`
Afterwards, it will be
`from dianna.utils import SimpleModelRunner`

I do think the second option is cleaner, and I agree we want to expose this to the user so let's merge it! (After fixing the import order)

Issue #91, opened 2021-12-27
Refactor Rise class into 2 (sub)classes


cwmeijer:
The Rise class contains code for 2 functionalities: explain text and explain images. Most code is exclusive for one or the other. There is little code that's really shared (maybe even none) and there is even some code duplication.
One possible solution for this would be to make some RISE superclass containing some shared logic, and subclasses for images and text. We then run into the following problem with the current implementation: We match the XAI method string with a class name, (for instance XAI method 'rise' matches class RISE). We won't be able to do that anymore when using multiple RISE classes. Also, while initializing the RISE object, we don't know yet whether the user is going to explain images or text with it.
Maybe the 'Factory' design pattern could help here, but still many implementation decisions need to be made. Post your suggestions or opinions in the comments please :-)

loostrum:
Closed via #325 
No inheritance was used, there is very little code duplications as the corresponding methods for text/images were already slightly different.

PR #325, opened 2022-09-06
Split RISE into separate classes for text and images (#91)

loostrum:


loostrum:
It seems to be working fine for RISE. Should we split the other methods as well in this PR? Now there's some boilerplate needed in __init__.py because not all methods are split.

loostrum:
Applied the split to LIME and KernelSHAP as well.

gcroci2 (approved):
Great! Seems very handy to have two separated class per data modality. Now adding functionalities to each method for each data modality will be much less complex.

egpbos:
Excellent!

PR #327, opened 2022-09-07
Fix text visualization in dashboard

loostrum:
When working on the tokenizers we changed how text visualization works, but didn't realize that the dashboard uses its own visualization function. This PR changes the dashboard visualization for text to work in the same way as in the main code


gcroci2 (approved):
Thanks! I have also run the dashboard locally using the code in this branch, and everything works well.

Issue #81, opened 2021-12-16
Compare without tolerance in tests


egpbos:
I see in the method tests that comparisons are done with a pretty high tolerance, e.g. here: https://github.com/dianna-ai/dianna/blob/5aabfc9e247494e212e1f85d91e75a085e3d7aff/tests/test_lime.py#L22

This makes the comparison very lenient, which may cause us to miss changes and/or bugs.

Is it possible to change them to exactly equal tests?

loostrum:
I think there'll always be floating-point issues so we can't check for exact equality, but we could decrease the threshold.
Then we should double-check that random seeds etc. are properly set.

cwmeijer:
discussion during the standup:
do we test whether it still looks like our expected heatmap? In that case we should keep a high value for atol.
Or do we want to check if anything has changed? in that case we should have a low value for atol.
In the standup we were all ok with either so no need to change, but you can if you like :-)

elboyran:
Don't remember what we decided at that spandup 10 months ago, but maybe, we can look at it while working on issue #80.

loostrum:
The decision is in @cwmeijer's comment: we didn't really care back then, but the two options (low or high tol) means we're testing different things. Let's decide in today's standup.

cpranav93:
We have picked a tolerance of 10^-5 (after a short vote)

egpbos:
Completely reasonable. Note that particle physicists use 10^-5 as well in typical LHC analyses as their error threshold when fitting parameters to data. So if you want to give the decision a more scientific sauce, you can cite them ;)

Issue #80, opened 2021-12-16
Compare complete saliency map to expected one in RISE tests on images


egpbos:
All other method tests compare the complete explanation result to a stored one except the RISE test on images, which only compares output shapes.

Standup label: discuss whether to do in this sprint or later.

cwmeijer:
standup conclusion: we should add this to the tests. Not in sprint 12 though because there's not enough time.

elboyran:
Expected LIME heatmaps should be added to the [test data](https://github.com/dianna-ai/dianna/tree/main/tests/test_data) and then compared to the RISE result on images in [test_rise.py](https://github.com/dianna-ai/dianna/blob/main/tests/test_rise.py).
See test_lime_function and test_rise_function in [test_lime.py](https://github.com/dianna-ai/dianna/blob/main/tests/test_lime.py)

loostrum:
You mean that RISE heatmaps should be added, right? We can't compare LIME
heatmaps to RISE results.
I'll have a look at adding these

elboyran:
@loostrum yes, add RISE heatmaps to be reference for comparison in the tests.

PR #373, opened 2022-10-19
Add reference heatmaps for RISE tests, change comparison tolerance to 1e-5

loostrum:
Closes both #80 and #81.


elboyran (approved):
Seems all fine.

PR #374, opened 2022-10-19
Avoid use of general exception

loostrum:
This PR introduces an `XAIMethodException`, to avoid the use of the generic `Exception` in `__init__.py`.
This makes SonarCloud happy


cwmeijer (approved):
A happy SonarCloud makes me happy ⛅

PR #377, opened 2022-10-25
Make user input for method no longer case sensitive

### LINKED ISSUE #375: Can we make the method selection not case sensitive?
The method name is given as string but needs to be converted to the right case to be able to import the related class.
The generated class name is method name + data modality (Image or Text), where the method name has the same case as whatever the user put in. This leads to some strange errors when the casing is wrong, as DIANNA then wrongly assumes the data modality is to blame. E.g.:
 ```
>>> dianna.explain_image(0, 0, "lime")
Traceback (most recent call last):
  File "/Users/loostrum/projects/DIANNA/dianna/dianna/__init__.py", line 95, in _get_explainer
    method_class = getattr(method_submodule, f"{method}{modality}")
AttributeError: module 'dianna.methods.lime' has no attribute 'limeImage'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/loostrum/projects/DIANNA/dianna/dianna/__init__.py", line 58, in explain_image
    explainer = _get_explainer(method, kwargs, modality="Image")
  File "/Users/loostrum/projects/DIANNA/dianna/dianna/__init__.py", line 97, in _get_explainer
    raise XAIMethodException(f"Data modality {modality} is not available for method {method}") from err
dianna.XAIMethodException: Data modality Image is not available for method lime
```

Suggestion: Make the method part of all class names uppercase (`KernelSHAPImage` -> (`KERNELSHAPImage`). Then we can just call `method.upper()` on the user's input.

loostrum:
Closes #375

### DIFF
--- a/dianna/__init__.py
+++ b/dianna/__init__.py
@@ -48,7 +48,7 @@ def explain_image(model_or_function, input_data, method, labels=(1,), **kwargs):
         One heatmap (2D array) per class.
 
     """
-    if method == "KernelSHAP":
+    if method.upper() == "KERNELSHAP":
         # To avoid Access Violation on Windows with SHAP:
         from onnx_tf.backend import prepare  # pylint: disable=import-outside-toplevel,unused-import
     explainer = _get_explainer(method, kwargs, modality="Image")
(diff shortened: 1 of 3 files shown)


cwmeijer (approved):
A very welcome addition to our user friendliness. Misspelling KernelSHAP and figuring out what was wrong already gave head aches while preparing the workshop last September. Great job!

Also, while googling I stumbled upon the casefold method. str.casefold() which I thought was interesting. Not applicable to our case though.

https://docs.python.org/3/library/stdtypes.html#str.casefold says:
    Return a casefolded copy of the string. Casefolded strings may be used for caseless matching.

    Casefolding is similar to lowercasing but more aggressive because it is intended to remove all case distinctions in a string. For example, the German lowercase letter 'ß' is equivalent to "ss". Since it is already lowercase, [lower()](https://docs.python.org/3/library/stdtypes.html#str.lower) would do nothing to 'ß'; [casefold()](https://docs.python.org/3/library/stdtypes.html#str.casefold) converts it to "ss".

    The casefolding algorithm is described in section 3.13 of the Unicode Standard.

    New in version 3.3.

Issue #357, opened 2022-09-27
Remove dependency on torchtext for dianna


cwmeijer:
Importing dianna now requires torchtext. I think we should somehow make this optional. Dianna shouldn't require installing TF or torch components. We could try to get rid of this dependency for instance by not importing tokenizers in the misc module. 

```
PS C:\Users\ChristiaanMeijer\Documents\robot_or_brain\robot_or_brain> python
Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import dianna
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\ChristiaanMeijer\AppData\Roaming\Python\Python38\site-packages\dianna\__init__.py", line 26, in <module>
    from . import utils
  File "C:\Users\ChristiaanMeijer\AppData\Roaming\Python\Python38\site-packages\dianna\utils\__init__.py", line 8, in <module>
    from .tokenizers import SpacyTokenizer
  File "C:\Users\ChristiaanMeijer\AppData\Roaming\Python\Python38\site-packages\dianna\utils\tokenizers.py", line 4, in <module>
    from torchtext.data import get_tokenizer
ModuleNotFoundError: No module named 'torchtext'
```

loostrum:
Agreed with removing the hard requirement on torchtext. Perhaps another variant may be of use here? dianna[text]?
Although too many variants would be confusing as well.

I'm not sure we can remove the tensorflow requirements, I think KernelSHAP needs it (which is also why that method can only run with an input onnx file and not a function)

PR #379, opened 2022-10-25
Remove torchtext dependency

loostrum:
The previous release is broken if the user does not have torchtext installed, see #357.
This PR makes torchtext optional, by not always importing the tokenizer anymore. An informative error is raised when the user tries to import a tokenizer when torchtext is not available. A dianna[text] variant is added to allow the user to automatically install the required dependencies for text-based explanations.

### DIFF
--- a/dianna/utils/__init__.py
+++ b/dianna/utils/__init__.py
@@ -5,4 +5,3 @@ from .misc import move_axis
 from .misc import onnx_model_node_loader
 from .misc import to_xarray
 from .onnx_runner import SimpleModelRunner
-from .tokenizers import SpacyTokenizer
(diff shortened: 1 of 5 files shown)


elboyran (approved):
Nice work!

cwmeijer:
It says, merging is blocked because not all commits were signed. @loostrum, do you know how to handle this? If so, we would like to know as well ;-) Otherwise we could nodge patrick to see if he knows.

loostrum:
Hmm I'll try following the steps for automatically signing commits with an SSH key (see here https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification), but I don't think that will help for already existing commits so I'll just merge this PR too. Let's see if the signing issues pops up again in the future.

PR #407, opened 2022-11-22
Make axis labels optional

### LINKED ISSUE #222: Can we make axis_labels optional?
While working on a "getting started" basic tutorial, I was thinking the "basic" usage is already pretty involved. One of the things you really have to explain to people is the `axis_labels` argument.

I'm a bit torn on this, because I think it's a very helpful feature, but in an ideal world, it should just be part of the input data, i.e. we should just require xarray or otherwise annotated arrays. That is not realistic, of course, so I'm trying to come up with alternative ways of making the `axis_labels` argument optional, while still promoting its use.

Ideas welcome!

### LINKED ISSUE #320: Input data without axis_labels gives cryptic exception message
When you try to explain an image and don't provide `axis_labels`, the user gets a message `AssertionError: Required label missing: channels`. This message could be a lot clearer for the user, explicitly suggesting what to do, namely: adding an `axis_labels={0:'channels'}`.

Related to #222 I think.

loostrum:
This PR is an attempt at making the axis_labels argument optional. The only required label left was the colour channels index. A new utils function is added that attempts to automatically determine the location of the channels axis. This does require a few assumptions:
- The channels axis has a size of either 3 (RGB) or 1 (greyscale)
- The channels axis is either the first or the last axis

When we can't find the channels axis, we raise an error and tell the user to use axis_labels.
Even if we do find it, we give a warning as we have no way of knowing for sure we correctly identified the channels axis.
I'm not sure always giving a warning is the best solution, other suggestions are more than welcome.

Note: I've also modified a few tests to make sure that for each method, there is at least one test that uses new functionality.

Some related issues that should be closed by this:
Closes #222 
Closes #320 (because we no longer use the required_labels option, so the user will not run into this error message anymore).

### DIFF
--- a/dianna/methods/kernelshap.py
+++ b/dianna/methods/kernelshap.py
@@ -151,7 +151,7 @@ class KERNELSHAPImage:
         axis_label_names = self.axis_labels.values() if isinstance(self.axis_labels, dict) else self.axis_labels
         if not axis_label_names:
             channels_axis_index = utils.locate_channels_axis(input_data.shape)
-            self.axis_labels = {'channels': channels_axis_index}
+            self.axis_labels = {channels_axis_index: 'channels'}
         elif 'channels' not in axis_label_names:
             raise ValueError("When providing axis_labels it is required to provide the location"
                              " of the channels axis")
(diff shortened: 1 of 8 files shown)


loostrum:
Possible addition: we could update some of our tutorials to remove the axis_labels argument.
This was actually one of the reasons to make it an optional argument. I'll leave it up to the reviewer of this PR to determine wether or not that should be part of this PR.

elboyran (approved):
Looks good. Only coverage complains. After merging should be OK (estimation).

PR #416, opened 2022-11-28
386 dashboard selection method for xai method

laurasootes:
This pull request creates a checklist for selecting the XAI method instead of a dropdown


cpranav93:
It looks much better as a checklist than it did as a drop-down! Great work @laurasootes. A couple of comments:
1) I think we no longer need to have 'import dash_bootstrap_components as dbc' in layouts.py. This can be removed. 

2) I see that the error message of "Missing image/model" is no longer raised under any circumstance. Is this an intentional change in the design?

A suggestion:
Could we have the three methods placed horizontally (side by side) instead of vertically? I think it might look better.

laurasootes:
Hi @cpranav93,
I fixed the first problem and your suggestion. I do indeed see that the error message is not longer accessible. It simply does not update anymore if you do not follow the correct order (i.e. if you have for example not uploaded an image). However, I think this is something that will be automatically fixed in your PR on the update button right? In that case I think it is not worth it to try to fix this here

dashboard/layouts.py
```diff
@@ -252,19 +252,20 @@ def get_uploads_images():
...
-                    dcc.Dropdown(id = 'method_sel_img',
+                    dcc.Checklist(id = 'method_sel_img',
                         options = [{'label': 'RISE', 'value': 'RISE'},
                                 {'label': 'KernelSHAP', 'value': 'KernelSHAP'},
                                 {'label': 'LIME', 'value': 'LIME'}],
-                        placeholder = "Select methods",
-                        value=[""],
-                        multi = True,
+                        inline=True,
+                        inputStyle={"margin-left": "10px", "margin-right": "5px"},
```
cpranav93, line 264:
```suggestion
                        inputStyle={"margin-left": "50px", "margin-right": "5px"},
```

cpranav93:
Hey @laurasootes, I think it works great. Added a comment regarding the spacing. Consider it, and whatever you decide, from my side the PR is ready to merge.

PR #426, opened 2022-12-06
make label a required argument (Fixes #131)

### LINKED ISSUE #131: Make `label` a required argument
Now the `label` argument is a keyword argument with a default of either 0 or 1, depending on the method.
Doesn't it make more sense to always ask the user which class they want an explanation for instead of assuming one?

cwmeijer:
I changed the signatures of all explain_image or explain_text definitions. I changed the label kwarg into a positional argument. No need for error checking because it is now automatically a TypeError to call any of those methods without a labels argument.

### DIFF
--- a/dashboard/callbacks.py
+++ b/dashboard/callbacks.py
@@ -163,10 +163,10 @@ def global_store_i(method_sel, model_path, image_test, labels=list(range(2)),
     elif method_sel == "KernelSHAP":
         relevances = dianna.explain_image(
             model_path, image_test,
+            labels=labels,
             method=method_sel, nsamples=n_samples,
             background=background, n_segments=n_segments, sigma=sigma,
             axis_labels=axis_labels)
-
     else:
         relevances = dianna.explain_image(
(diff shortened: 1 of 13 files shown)


loostrum:
The error checking is at least working as advertised, I see a whole bunch of TypeErrors in the tests :D

cwmeijer:
Always great rushing commit out 1 minute before going home ;-) Back into draft it is.

cwmeijer:
Everything should be OK now.
@loostrum could you have another look?

loostrum (commented):
Looks good! I don't understand why the fair-software CI is failing, it seems unrelated to this PR. 
The linter is complaining about some dashboard things that mostly seem unrelated as well, but there is one about the labels argument you may want to check out before merging, see https://github.com/dianna-ai/dianna/actions/runs/3655705906/jobs/6177325029#step:4:13

loostrum (approved):

elboyran:
Signed commits issue. @cwmeijer would you know how to merge this one?
Also linting errors.

PR #432, opened 2022-12-20
added center generator

NicoRenaud:
This PR adds the possibility to add a generator to the mask creation. The generator yield a new center where the mask should obscure the features. In the case of deeprank this generator can be used to create masks that predominantly mask region of interests of the protein-protein interface


loostrum (approved):
Very nice, thanks a lot for this Nico!

Issue #437, opened 2023-01-09
Lime explanation crashes sometimes - special character related


laurasootes:
Relevances for Lime cannot always be calculated. The origin lies in that the created sentences are not always returned in equal length

loostrum:
After quite some investigation we found that the issue lies in how LIME creates a list from the tokens. Take this example:
```'This movie is great !?'```
Tokenized, this would be:
```['This', 'movie', 'is', 'great', '!', '?']```
Now when LIME creates a list that represents the original input, it uses the same separators as the original, so like this:
```['This', ' ', 'movie', ' ', 'is', ' ', 'great', ' ', '!', '?']```

Let's say both special characters are masked and replaced by "UNKWORDZ" (the lime default).
The final sentences will then be
```This movie is great UNKWORDZUNKWORDZ```
The is no space between the two UNKWORDZ instances. When tokenizing this again, the resulting list is one token shorter than the original tokenized text. This mismatch in length of the sentence results in the "Cannot convert input to tensor" error. 

The proposed solution is to attempt to fix this in DIANNA by detecting special characters in the input and adding spaces around them, before forwarding the input to LIME. We have to be careful with the apostrophe: the input `let's` should be tokenized as `["let", "'s"]`.

A possibly incomplete list of special characters we might want to handel (taken from the vocab in the movie reviews model runner):
```
,
.
-
`
--
'
...
)
(
''
:
;
/
?
*
!
$
&
#
+
=
```

loostrum:
I'll try implementing this in the SpacyTokenizer class that we have. That way it doesn't interfere with other users, that will probably user their own tokenizers.

loostrum:
Unexpected situation: I tried adding a test to run with special chars, expecting it would fail and I could then fix it, but the test passed. Turns out that the movie review model runner as defined in tests/utils.py _does_ work, while the one in the dashboard/utilities.py does not work. I'll check the differences between the two and fix the one in the dashboard

loostrum:
The difference is that the model runner in the dashboard runs all sentences through the model at once, while the one in tests does it one by one. Running it one by one hides the error rather than actually fix it.

In general it would make sense to run the sentences one by one so we can support running several different reviews of different length at the same time, but in reality we're always passing a single review. So I suggest we put the model runner that runs all sentences at the same time in both the dashboard and tests. That will break the test, which we can then fix by adding the special character support in the SpacyTokenizer class.

Issue #435, opened 2023-01-09
Long text does not fit in dashboard explanation image


laurasootes:
In the text tab of the dashboard, the explanation of the text printed as an image that is created as a loaded screenshot:

```
output = _create_html(input_tokens, relevances_lime[0],
                        max_opacity=0.8)
hti = Html2Image()
expl_path = 'text_expl.jpg'
hti.screenshot(output, save_as=expl_path)
im = Image.open(expl_path)
im = np.asarray(im).astype(np.float32)
```

However, if the text is long it does not fit in the image. This seems to be an issue with the screenshot and not with the width of im. 

![image](https://user-images.githubusercontent.com/95367983/211289819-06af95b6-9c9f-43db-abc8-ac6f8f5ff3d7.png)

PR #438, opened 2023-01-10
429 implement recent dashboard layout changes to text tab

laurasootes:
I added/translated the changes made to the text page of the dashboard (update and stop button, selection of model, parameters of model).
Everything seems to work, except for the issues that I created (#435 and #437 ), which are not problems in the translation but merely problems that already existed in the text page that we did not know.

This branches also fixes the linting error about the tensor flow version


cpranav93:
Hey @laurasootes, could you provide me with the model for testing the text explanations? I currently only have the onnx model for the images. Thanks!

laurasootes:
The model is already in the repo under app_data. It is the movie review model.

cpranav93:
Hey @laurasootes, I am trying out the dashboard>text and no matter what text I give, I get an error (see attached screenshot). I am not sure if I am doing something wrong here.

image

laurasootes:
but can you find where and why it is going wrong in the code?

cpranav93:
I had a look at where the error was happening and it is in line 546: 
hti = Html2Image()

I do not know enough about dash to figure out yet why this is causing an exception which gives the error message.

Also, I noticed that when I refresh the page, and input just the text and model without the method, it raises
'There was an error running the model. Check either the test text or the model.' 
but it should raise:
'Missing model, input text or XAI method.'

This can be fixed in line 489 by changing it to:
if (fn_m and input_text and sel_methods) is not None:

dashboard/layouts.py
```diff
@@ -650,29 +785,29 @@ def get_uploads_text():
...
-                    id='graph_text_rise',
+                    id='graph_text_lime',
                     figure = utilities.blank_fig())],
                     className = 'row',
                     style = {
                         'margin-top': '80',
                         'margin-left': '140px',
                         'height': '100px'
-                    })
+                    })
```
loostrum, line 810: The extra blank space at the end is not needed here :)

loostrum (approved):

Issue #423, opened 2022-12-01
Linting CI is broken


loostrum:
See the following screenshot from https://github.com/dianna-ai/dianna/actions/runs/3593771040/jobs/6051161094:

![image](https://user-images.githubusercontent.com/6370787/205088266-35901e92-7ee6-4509-a9c6-401a4473ff2d.png)

loostrum:
We already install prospector[with_pyroma] exactly as suggested in the screenshot and it worked before. Perhaps an update to prospector/pyroma broke this

egpbos:
Highly likely. There was an update yesterday: https://pypi.org/project/prospector/#history Also an update three hours ago, so maybe it was fixed. But looking at the log of that run, I also see this:

<img width="564" alt="Schermafbeelding 2022-12-02 om 16 22 23" src="https://user-images.githubusercontent.com/6146598/205326501-30f67b26-439c-4c64-be61-91cf9e2e506c.png">

Not sure if 1.8.2 fixes that...

egpbos:
Maybe just try installing pyroma explicitly? Or disable it in linting config, not sure what it does or whether it is even used here anyway (it gives a score, not a fail/pass, I think).

loostrum:
Ah good catch. There's already an open issue in the prospector repo: https://github.com/PyCQA/prospector/issues/545
It can still fail even if pyroma is installed, so we could also just pin the prospector version to 1.8.0 for now

loostrum:
I'm also not sure what pyroma actually does. I opened a PR with the least-effort way of getting it to run again: pinning the version

PR #442, opened 2023-01-11
Unpin prospector

egpbos:
Version 1.8.4 should have fixed the issues previously mentioned in #423. See https://github.com/PyCQA/prospector/milestone/7?closed=1. Also relevant to the discussion in https://github.com/NLeSC/python-template/issues/327

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,8 +62,8 @@ install_requires =
 [options.extras_require]
 dev =
     bump2version
-    prospector[with_pyroma]==1.7.7  # last version that works, see https://github.com/PyCQA/prospector/issues/545
-    pylint==2.15.6
+    prospector[with_pyroma]
+    pylint
     isort
     pytest
     pytest-cov


egpbos:
The linter runs, so I think this should be considered an improvement over pinning the old version? The failures are due to actual linter suggestions, which I guess are a separate issue :)

loostrum (approved):
Nice, good that the linter is running again. I agree with unpinning prospector and creating a separate issue for the linting problems. The pylint version pin was unrelated, that was already there before the prospect pin. I don't know why and I'm tempted to just unpin pylint as well, because I don't see why it was pinned in the first place.

egpbos:
Oh right, I didn't check if pylint was necessary to unpin. Is that then causing (some of) the linting failures now? Anyway, may indeed be good to unpin as well, especially if there's no reason to keep it pinned, but maybe someone else knows why it was pinned, @cwmeijer or @elboyran?

cwmeijer:
I don't remember anything about pinning prospector. It doesn't seem sensible to me to pin the version of a well maintained linter package. Unpin it :-)

PR #450, opened 2023-01-18
refactor rise and fix linter issues (refs 447)

cwmeijer:
This solves all linter issues currently in main.

Most issues were solved by adding disabling comments. Wherever possible, I tried adapting the code instead to satisfy to linter rules. For instance, I did some refactoring to get the number of variables in scope down in `rise.py`.


cpranav93 (approved):
All good!

Issue #400, opened 2022-11-22
Implement label file upload support in dashboard for images


cwmeijer:

elboyran:
@laurasootes I guess you'd be most interested in reviewing this one?

PR #451, opened 2023-01-23
400 implement label file upload support

cpranav93:
This PR address issue #400 for implementing label upload support in the dashboard. A new box is added for submitting labels in a .txt file. There is currently no checking of the labels and it assumes that the user is uploading it accurately. The only checks are for whether the file is in the correct format and if it is empty or not.


cpranav93:
@laurasootes, it seems that the size of the box varies with the size of the window. 
image

laurasootes:
Ah yeah true, I see it now on my larger screen

PR #452, opened 2023-01-24
441 implement text side by side

laurasootes:
Created:
- second text upload option
- Input parameters are the same for both strings
- Predictions are made and printed for each of the strings
- Relevances are plotted side by side for the different strings

Should work for one or two selected methods, and one (does not matter which one) ore two uploaded strings.


cpranav93 (approved):
Everything looks great! Go ahead and merge.

I somehow still cannot get the text dashboard to work properly. I should fix it...

PR #482, opened 2023-02-21
Cache python install in github actions

### LINKED ISSUE #475: Caching for github actions
I just noticed that your tests are taking a very long time to run. They can be much, much faster if you cache the pip install step.

Can be done by adding this step to the actions:

```yaml
      - uses: actions/cache@v3
        with:
          path: ${{ env.pythonLocation }}
          key: ${{ env.pythonLocation }}-${{ hashFiles('setup.cfg') }}
```

stefsmeets:
This PR caches the python environment. This will significantly reduce the time it will take to run the tests on github actions.

All steps should run roughly 2-3x faster now. I also made it so that the tests only run when the PR is ready for review. This helps save cpu cycles, and saving cpu cycles is good for the environment 🌱

Closes #475

### DIFF
--- a/.github/actions/install-python-and-package/action.yml
+++ b/.github/actions/install-python-and-package/action.yml
@@ -13,7 +13,7 @@ runs:
   using: "composite"
   steps:
     - name: Set up Python ${{ inputs.python-version }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ inputs.python-version }}
     
--- a/.github/actions/install-python-and-package/action.yml
+++ b/.github/actions/install-python-and-package/action.yml
@@ -16,13 +16,18 @@ runs:
       uses: actions/setup-python@v3
       with:
         python-version: ${{ inputs.python-version }}
-        cache: 'pip'
-        cache-dependency-path: setup.cfg
+    
+    - uses: actions/cache@v3
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.pythonLocation }}-${{ hashFiles('setup.cfg') }}
+    
     - name: Python info
       shell: bash {0}
       run: |
         which python3
         python3 --version
+    
     - name: Upgrade pip and install dependencies
       shell: bash {0}
       run: |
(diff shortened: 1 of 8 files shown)


stefsmeets:
Hi @cwmeijer Could you have a look at this PR?

.github/actions/install-python-and-package/action.yml
```diff
@@ -13,18 +13,23 @@ runs:
...
       run: |
         which python3
         python3 --version
+    
     - name: Upgrade pip and install dependencies
       shell: bash {0}
       run: |
         python3 -m pip install --upgrade pip setuptools wheel
-        python3 -m pip install .[${{ inputs.extras-require }}]
+        python3 -m pip install -e .[${{ inputs.extras-require }}]
```
cwmeijer, line 35: Fine either way of course. Is there any advantage of installing 'editable' like this in this case as part of CI?

.github/workflows/build.yml
```diff
@@ -5,6 +5,11 @@ on:
...
     branches:
     - main
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
   pull_request:
     branches:
     - main
```
cwmeijer, line 15: I haven't worked a lot with filters and trigger types etc but this doesn't look how I'd expect it to.
To me this reads like:
We run the workflow if anything is pushed to main, and I'm not sure how to interpret these triggers in this case.
We also run the workflow on PRs when their base is main (using the default triggers as nothing is specified under this).

In other words, shouldn't the `types` element and it's children be under `pull_request` instead of under `push`?

cwmeijer (approved):
Great idea and great improvement 🍀
I have some questions in the comments. Could you please check the location of 'types' within the yaml? If you are sure it's fine as it is, it would be great if you could explain it to me :-)
Thanks for the PR!

.github/actions/install-python-and-package/action.yml
stefsmeets: I did this to avoid caching the dianna install.

.github/workflows/build.yml
stefsmeets: Yep, good point, I will fix this later today 👍

Issue #490, opened 2023-02-28
Self implement lime methodology for time-series


cpranav93:
The existing methods for lime (lime for time/ limesegment) do not have licenses and also appear to not work very well. Refer to #459.

Therefore, we would like to implement our own version of the algorithm for dianna.

Issue #476, opened 2023-02-21
Decide on time series format support


cwmeijer:

elboyran:
@stefsmeets , would this be something you'd be interested in doing/contributing to?

cwmeijer:
we chose a 2d np.array with shape (timesteps, channels)

PR #491, opened 2023-02-28
Add visualization for timeseries

### LINKED ISSUE #457: Define visualization for time series data domain
Look at existing visualizations in other packages
what about different research domains. How are time series visualized there.
Make a fine design for our visualization

stefsmeets:
This PR implements a visualization for timeseries visualization. I tried to implement what seems to be the most common way of making these plots.

Currently it takes `x`, `y` and a bunch of segments that look like this:

```python
from dianna.visualization import plot_timeseries

segments = [
    {'index': 0, 'start': 0, 'stop': 2, 'weight':-0.6},
    {'index': 1, 'start': 2, 'stop': 4, 'weight':-0.3},
    {'index': 2, 'start': 4, 'stop': 6, 'weight': 0.0},
    {'index': 3, 'start': 6, 'stop': 8, 'weight': 0.4},
    {'index': 4, 'start': 8, 'stop': 10, 'weight': 0.7},
]

x = np.linspace(0, 10, 20)
y = np.sin(x)

plot_timeseries(x=x, y=y, segments=segments, cmap='bwr', show_plot=True)
```

This is what it looks like:

Screenshot 2023-02-28 162944

I tried to keep the interface somewhat basic, because I expect this will need some finetuning after #476 and #490

Closes #457


stefsmeets:
@cpranav93 Could you have a look and let me know what you think?

dianna/visualization/timeseries.py
```diff
@@ -0,0 +1,70 @@
...
+def plot_timeseries(
+    x: np.ndarray,
+    y: np.ndarray,
+    segments: List[Dict[str, Any]],
+    xlabel='x',
+    ylabel='y',
+    cmap: Optional[str] = None,
+    show_plot: bool = False,
+    output_filename: Optional[str] = None,
+    ax: Optional[plt.Axes] = None,
```
cwmeijer, line 16: Thanks for the type hints! We should do that more in DIANNA.

cwmeijer (approved):
Looks great and I also like the input format. I think it is perfect.
I don't suppose this works on multivariate time series right? That is totally fine, but I think the function title (or at least the doc string) should mention this.
Also, please fix the linting before merging. :-)
Great addition! 🎆

cpranav93 (approved):
Looks great! As we expand the time-series functionality in Dianna we can add more methods for visualisation if necessary. Prospector seems to be failing for the build - maybe worth a look before merging.

stefsmeets:
Oh, we are using pylint 😬 I will address this and then merge. Thanks!

```
dianna/visualization/timeseries.py
  Line: 7
    pylint: too-many-arguments / Too many arguments (9/5)
    pylint: too-many-locals / Too many local variables (16/15)

tests/test_visualization.py
  Line: 5
    pylint: missing-function-docstring / Missing function or method docstring
  Line: 18
    pylint: trailing-newlines / Trailing newlines
```

stefsmeets:
@cwmeijer Multiviarate timeseries can be easily added later. I would suggest to add that once it is more clear what data needs to be plotted.

cwmeijer:
I absolutely agree with adding multivariate functionality later. Please still update the doc string because who knows how many releases we will do before actually adding multivariate support.

PR #495, opened 2023-03-03
Added label file upload for text

cpranav93:
@laurasootes, I have added label file upload support to text as well. However, I am not able to test it due to WSL issues for text. Could you give it a go and let me know if it works fine?

### DIFF
--- a/dashboard/callbacks.py
+++ b/dashboard/callbacks.py
@@ -454,7 +454,6 @@ def upload_label_text(filename):
                     lnames = f.readlines()
                 
                 labelnames = [item.rstrip() for item in lnames]
-                print (labelnames)
                 if labelnames is None or labelnames == ['']:
                     return html.Div(['Label file is empty, please upload a valid file']), labelnames
                 return html.Div([f'{filename} uploaded']), labelnames
--- a/dashboard/callbacks.py
+++ b/dashboard/callbacks.py
@@ -25,7 +25,6 @@ from dianna.utils.tokenizers import SpacyTokenizer
 import utilities
 from utilities import MovieReviewsModelRunner
 from utilities import _create_html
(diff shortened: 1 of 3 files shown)


dashboard/callbacks.py
```diff
@@ -51,9 +50,9 @@
 cache.clear()
 
 # global variables # replace by generic label loader
-class_name_mnist = ['digit 0', 'digit 1']
-class_name_text = ["negative", "positive"]
-class_names_imagenet = [imagenet_class_name(idx) for idx in range(1000)]
+# class_name_mnist = ['digit 0', 'digit 1']
```
laurasootes, line 53: these lines can be removed instead of commented

laurasootes: additionally, I just noticed that there is a print statement in line 158 from the image label update

laurasootes (commented):
I think it is not working yet, somehow it is using the second character (can even be a space) of the input string as prediction label:

image

image

dashboard/callbacks.py
cpranav93: Comments and print statement removed.

cpranav93:
The label issue should be fixed now. Could you give it a test? @laurasootes

dashboard/callbacks.py
```diff
@@ -537,9 +559,10 @@ def update_multi_options_t(fn_m, input_text, sel_methods, new_model, new_text,
             onnx_model_path, word_vector_path, max_filter_size=5)
 
         try:
+            print ('label:', labelnames)
```
laurasootes, line 562: Can you remove this print statement

laurasootes (approved):
Everything works and looks good! If you can remove this one redundant print statement it is ready to go :)

PR #497, opened 2023-03-03
467 write a short readme on how to start the dashboard

laurasootes:
I wrote a short readme for the usage of the dashboard. Please let me know if this is what you had in mind, or if it needs to be extended

Issue #414, opened 2022-11-24
Dashboard: create collapse menu for parameters


laurasootes:
Maybe a nice to have:

create a dropdown menu in the dashboard for the model parameters:
https://dash-bootstrap-components.opensource.faculty.ai/docs/components/collapse/

(had a quick look at it, but noticed a potential issue with required stylesheet to have it actually collapse
`dash.Dash(external_stylesheets=[dbc.themes.BOOTSTRAP]))`

stefsmeets:
`dash.Dash(external_stylesheets=[dbc.themes.BOOTSTRAP]))`

This indeed updates the current styling and messes up the layout to some degree.

Issue #493, opened 2023-03-01
pylint configuration


stefsmeets:
I have also noticed the anti-pattern that that many of the issues pylint checks for against are ignored by pragmas. Searching for `# pylint` gives 36 results, mainly related to imports, too many arguments/locals and protected variable access:

- `import-error`
- `import-outside-toplevel`
- `protected-access`
- `too-many-arguments`
- `too-many-locals`
- `unused-argument`
- `unused-import`

My suggestion would be to re-configure pylint and simply ignore these warnings altogether (or actually fixing the underlying issue).

Issue #492, opened 2023-03-01
Code formatting and pre-commit hooks


stefsmeets:
Dianna runs a fairly strict pylint configuration that is somewhat complainy. It's also slow, which makes it a bit of a chore to run while doing the prospector/pylint dance. 

Many of the issues it raises are easily fixed by using code formatters like autopep8, yapf, or black. I would be happy to configure one of these (yapf or autopep8 would be my choice) for dianna for code consistency and saving my time.

I'm not in favour of running too many linters/formatters locally, like the documentation [recommends](https://github.com/dianna-ai/dianna/blob/main/docs/developer_info.rst#running-linters-locally). And the pre-commit hook hides the output and requires some manual steps to apply the fixes.

There is an excellent tool called [pre-commit](https://pre-commit.com/) that makes it very easy to configure different checks, with automated fixes and it only runs on the code that has been changed. If you agree, I will happily configure it for dianna to run one of the autoformatters above, as well as pylint/pycodestyle.

cwmeijer:
This would be great and will save us a lot of time for sure.

PR #502, opened 2023-03-07
Refactor dashboard

stefsmeets:
Trying to make sense of the dashboard for #414, I refactored the different elements to make it a bit easier to work with.


stefsmeets:
Would this be useful for the project? Could you have a look @laurasootes ?

laurasootes:
fixes https://github.com/dianna-ai/dianna/pull/497#issuecomment-1457888702

laurasootes (approved):
Thanks! Pulling apart these different parts already helps a lot. The images.py and text.py remain large and hard to read files, but I am not sure if/how we can make that better. I don't know if we should also try to pull apart some parts of the callbacks file

laurasootes:
Only thing: you made the linter unhappy :)

stefsmeets:
Thanks, the linter also makes *me* unhappy 😅
I will fix the issues and then merge 🚀

laurasootes:
@stefsmeets I do not want to make you unhappy again, but the linter still is 😥

stefsmeets:
This linting is madness. How do you guys work without auto-formatters that fix these automatically 😬

I can address this with #492 #493

laurasootes:
Sounds like a plan. The lining is currently broken ~90% of the time

PR #504, opened 2023-03-07
Add button to show/hide parameters

### LINKED ISSUE #414: Dashboard: create collapse menu for parameters

stefsmeets:
This PR adds a collapsible parameters section.

One issue is that this requires an external stylesheet that changes our layout slightly.

Follows up on #502 

Closes #414


stefsmeets:
Hi @laurasootes , could you have a look and let me know if this is what you had in mind?

laurasootes:
The collapsable parameter menu looks good! This exactly what we had in mind in terms of functionality.

However, this PR seems to have broken 3 things:
- the submit button reaches all the way to the side, there used to be some space here
- The select XXX is no longer light blue
- The update explanation buttons does not work in the text tab (does work in the image tab)

image

stefsmeets:
Yeah, this is a side-effect of adding another style sheet. It clashes with all the inline styling done in the dashboard.

laurasootes:
hmm, annoying. I presume it is fixable? 👼

laurasootes:
For the image tab the run fails and gives an "tuple index out of range" when trying run it using RISE (not for the other methods, also not in text tab)

stefsmeets:
> For the image tab the run fails and gives an "tuple index out of range" when trying run it using RISE (not for the other methods, also not in text tab)

Can you tell me your inputs? Because I don't see this happening on my side.
Can you also test if you see it in the main branch?

laurasootes:
Turns out it was me, so no problems

laurasootes (approved):

PR #506, opened 2023-03-09
478 rise for time series

cwmeijer:
Add RISE for timeseries.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.6.0
+current_version = 0.7.0
 
 [comment]
 comment = The contents of this file cannot be merged with that of setup.cfg until https://github.com/c4urself/bump2version/issues/185 is resolved
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -41,7 +41,7 @@ authors:
     orcid: "https://orcid.org/0000-0002-7896-2969"
 
 doi: 10.5281/zenodo.5801485
-version: "0.6.0"
+version: "0.7.0"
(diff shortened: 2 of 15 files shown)


cwmeijer:
I added RISE and 2 tests for it. I didn't test it with actual data or with an actual model. There is also something not right with the output format. It is inconsistent with the timeseries visualization. RISE is now returning a saliency map as an np array while the visualization is expecting a dict with starts and stops and weights. Therefore, the test_common_usage.py contains some transformation code.

@geek-yang, maybe you would like to have a look what should be a good general format for timeseries saliency. I suppose the dict option is not bad and quite general, but it contains an 'index' key which I don't understand why that's useful. Also, maybe creating a dataclass for this is better than dicts.

tests/methods/time_series_test_case.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+
+
+"""In this test case, every test instance is a 28 days by 1 channel array indicating the max temp on a day."""
```
geek-yang, line 4:
```suggestion
"""Test case for timeseries xai methods.
This test case is designed to show if the xai methods could provide reasonable results.
In this test case, every test instance is a 28 days by 1 channel array indicating the max temp on a day.
"""
```

geek-yang (approved):
Awesome work! I really like the idea about having the hot and cold day dummy data to test if the method works as expected or not. This solves our big puzzle and it can be applied to test other methods as well. Well done!

geek-yang:
Follow our discussion, I have created a new task on the board related to the notebook #539 . I will polish the notebook and add an ONNX example there.

PR #508, opened 2023-03-13
Add pre-commit for linting/auto-formatting

### LINKED ISSUE #492: Code formatting and pre-commit hooks

### LINKED ISSUE #493: pylint configuration

### LINKED ISSUE #496: isort import error
Run isort --check-only dianna --diff
ERROR: /home/runner/work/dianna/dianna/dianna/visualization/timeseries.py Imports are incorrectly sorted and/or formatted.
--- /home/runner/work/dianna/dianna/dianna/visualization/timeseries.py:before	2023-03-03 09:54:58.284623
+++ /home/runner/work/dianna/dianna/dianna/visualization/timeseries.py:after	2023-03-03 09:57:00.2[15](https://github.com/dianna-ai/dianna/actions/runs/4322243331/jobs/7544416158#step:5:16)377
@@ -1,5 +1,7 @@
-from typing import Any, Dict, List, Optional
-
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
 import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import cm
Error: Process completed with exit code 1.

stefsmeets:
This PR adds `pre-commit` as a tool for managing pre-commit hooks and code linting. It also replaces pylint by ruff as a linter. Ruff also sorts imports similar to `isort`. New checks can be enabled in `pyproject.toml`.

Running the linters now goes through [pre-commit](https://pre-commit.com/).

```
# staged files only
pre-commit

# all files
pre-commit run --all-files
```

You can enable automatic linting and code formatting with `pre-commit` on each commit:

```
pre-commit install
```

This replaces the git hook from before. The advantages are that this only checks changed files, issues are automatically fixed (if possibe/enabled), and it is much faster than prospector/pylint.

If you use the old hook, you will need to run:
```
git config --unset core.hooksPath
```

You can also install ruff locally (`pip install ruff`) and run it on the entire codebase: `ruff check dianna --fix`.

I also added `yapf` as a code formatter. At some point we should run yapf on the entire code base so that everyone is up-to-date. This may hurt a bit with open PRs, but the reward is that we no longer have to worry about code formatting / fighting with pylint. I will create a new PR once this one is accepted.

I realized I could not make this PR without fixing some standing issues to pass the linter, so I fixed some whitespace issues, added missing parameters in docstrings, and looked at all the pylint pragma's.

Closes #492 
Closes #493 
Closes #496


stefsmeets:
@cwmeijer Would you be willing to review this? Let me know what you think!

dianna/__init__.py
```diff
@@ -34,31 +33,30 @@
...
+        kwargs: These keyword parameters are passed on
 
     Returns:
         One heatmap (2D array) per class.
 
     """
     if method.upper() == "KERNELSHAP":
         # To avoid Access Violation on Windows with SHAP:
-        from onnx_tf.backend import prepare  # pylint: disable=import-outside-toplevel,unused-import
+        from onnx_tf.backend import prepare  # noqa: F401
```
cwmeijer, line 52: I would rather keep the human readable version. Is that something you can configure?
I'm surprised the linter instruction isn't removed at all for this line, because for instance in kernelshap and other places you are removing the ignore-unused-import instruction completely.

cwmeijer (changes requested):
Looks great!
I have to say that I don't fully understand what code is responsible for actually changing the code before a commit, but I'm sure you'll explain me soon ;-)
Could you have a look at my comment?

dianna/__init__.py
stefsmeets: Without the pragma, `ruff` replaces this entire import by `pass`, because the import is not used. This is what happened in all the other instances of `unused-import`.

For this instance I added back the noqa, because of the comment above. I guess there is some side effect from the import that dianna requires. 

I can't configure the noqa in a different way, this is the convention used by `ruff`.

stefsmeets:
Yeah, good question!

So what happens is that `pre-commit` checks which files have changed (essentially `git diff`). Every file that is changed gets passed to some hooks based on the file extension. For example, it does not make sense to pass a yaml or markdown file to a python linter.

Each of the hooks then runs on the changed files. These can then change the code if they wish. This is no different than running each of the hooks individually on the command line. Some simply emit warnings (like pylint), others also update the code. You will have to stage these again and if they pass without changes then the `pre-commit` hook is passed.

All these hooks are defined in: https://github.com/dianna-ai/dianna/blob/61fe210edb8c95e69e2b54bc6ea85cc3eb220c8e/.pre-commit-config.yaml

You'll notice there are some that check whether yaml files are valid (check-yaml), or whether the python code actually parses correcty (check-ast). Then there are the heavy lifters yapf and ruff which format and lint the code. I also tentatively added mypy, which we could enable at some point for type checking.

stefsmeets:
I guess we can enable mypy here as a step towards #85

dashboard/callbacks.py
```diff
@@ -176,11 +176,9 @@ def upload_label(filename):
 # these computations are cached in a globally available
 # redis memory store which is available across processes
 # and for all time.
-# pylint: disable=dangerous-default-value
-# pylint: disable=too-many-arguments
 @cache.memoize()
 def global_store_i(method_sel, model_path, image_test,
-                   labels=list(range(2)), axis_labels={2: 'channels'},
+                   labels=(0, 1), axis_labels={2: 'channels'},
```
egpbos, line 181: You can make it `[0, 1]` so that you needn't do the `labels = list(labels)` call below.

stefsmeets: It's dangerous to do that, and not recommended. A list is mutable and `labels` only gets initialized during import time. So if any function downstream updates the labels (i.e. `labels[0]='foo'`), your default arguments are now `labels=['foo', 1].

pyproject.toml
```diff
@@ -21,3 +21,77 @@ commands = pytest
...
+[tool.ruff]
+# Enable Pyflakes `E` and `F` codes by default.
+select = [
+    "F",  # Pyflakes
+    "E",  # pycodestyle (error)
+    "W",    # pycodestyle (warning)
+    # "C90",  # mccabe
+    "I",    # isort
+    "D",    # pydocstyle
+    # "PL",   # Pylint
```
egpbos, line 33: I would enable these as well. Pylint is complainy, and annoying, yes, but it is often right. I can imagine that within Ruff, only the most "right" ones are added anyway, so it's probably a good idea to have the Ruff-filtered Pylint complaints for sure.

The many "too many arguments" complaints that this PR removes, for instance: I think they make a lot of sense to think about. It's very easy to call functions with parameters in the wrong order. Mypy will help a bit with this, but if parameters have the same types, it doesn't. A potential fix is to use keyword-only arguments, but that makes your calls very long. An alternative is to use configuration objects, probably dicts or dataclasses. Dataclasses especially can be nice to pass along configuration between many functions in a consistent way, can check consistency internally with class methods, etc.

tests/test_kernelshap.py
```diff
@@ -16,7 +15,7 @@ def test_shap_segment_image(self):
         n_segments = 50
         compactness = 10.0
         sigma = 0
-        image_segments = explainer._segment_image(  # pylint: disable=protected-access
+        image_segments = explainer._segment_image(
```
egpbos, line 18: Here another very deserved complaint by pylint :) It's bad form to access "protected" functions. Sure, it is a test, in this case, but does it really make sense to have `_segment_image` protected? Usually, protected (or private with double underscore) functions are protected/private because they modify internal class/object properties that you don't want users to mess with. In this case that is totally not true: the function is static (so it changes nothing in objects) and in fact only passes along parameters to another, external function. There's no reason to keep it protected.

Note: I'm not trying to review the contents here, I'm just giving examples of pylint's justified complaining :D

dashboard/callbacks.py
egpbos: Ah great point, hadn't thought about that!

egpbos:
Super nice PR! Sorry for the unasked review, but I was just very interested to see how this would look in practice and in particular in DIANNA which I know well :) My only minor complaints are about disabling pylint complaints, as the examples above illustrate ;) I know there's always a trade-off with being realistic in terms of how much time you can spend on polishing. But DIANNA is growing into quite a big thing, so I would argue that in this case the strictest standards should be seriously considered.

pyproject.toml
stefsmeets: The pylint implementation of ruff is incomplete, which is why I have not enabled it. I'm afraid it will become too much of a moving target.

I agree that we should think about 'too many arguments', but if the solution is to slap on a `# pylint: too-many-arguments`, then this whole exercise is now a giant anti-pattern that we should try to avoid. It's also not necessarily bad to have too many arguments. The code still runs. Sometimes the work-arounds (i.e. passing around dataclasses and kwargs) are more harmful than what it tries to prevent.

egpbos: Yes true. I also don't like so much that you have to annotate the code to stop the complaints. Ideally, you'd have some central file where you list exceptions to strict checks. Then, perhaps, periodically, or when you are doing a major refactoring operation anyway, you can review all the warnings to easily find design-smells that perhaps you can fix in the same refactoring in one go. I still think it's good to check them, but maybe it could be a separate CI workflow as well. Probably, that one will be red most of the time then, but you can look at it every now and then.

cwmeijer (approved):

cwmeijer:
Sorry for the delay in approving. LGTM asap :-)

stefsmeets:
Cool, I will do a full re-format sometime next week.

Also a reminder to run:
```
git config --unset core.hooksPath
```
to remove the old hook.

Issue #512, opened 2023-03-14
Dashboard: add label file for RESnet


laurasootes:
add a txt label file for the 1000 labels of the RESnet model to the dashboard folder

PR #513, opened 2023-03-15
create label file for ResNet model

laurasootes:
Created label file for ResNet model, included script and resulting txt file


cpranav93 (approved):
It looks great! My only thought is: Should we move the get_resnet*.py to the tutorials folder or elsewhere? or is this the best place for it?

cpranav93:
Also good to fix the linter errors on the get_resnet.py file!

laurasootes:
yeah, I totally agree. This script is really only to show how I obtained the txt file and has no other purpose. Maybe I can simply paste the script in this issue for the logging, and then delete the file as it has no other use

cpranav93:
Yeah, that is a good idea! Good to have it archived somewhere (such as here) but its not needed otherwise.

laurasootes:
### Code used to generate label file

````python
"""write all class labels for the ResNet50 model to a txt file"""
import numpy as np
from tensorflow.keras.applications.resnet50 import decode_predictions

class_names = [decode_predictions(np.eye(1,1000, i))[0][0][1] for i in range(0, 1000)]

with open('labels_resnet50.txt', 'w', encoding="utf-8") as file:
    file.write('\n'.join(class_names))

laurasootes:
Closes #512

PR #533, opened 2023-03-28
Run CI only on PRs ready for review

### LINKED ISSUE #519: Fix github actions running when PR is in draft state
In #482 I intended to make a change so that the tests only run when the PR is marked as ready for review.  This saves CI use which in turn reduces the footprint of DIANNA 🌱

This does not happen at the moment, because I forgot to add the following line to the github actions:

```
jobs:
  build:
    name: ...
    if: github.event.pull_request.draft == false
    ...
```

stefsmeets:
This PR adds a line to only run the CI on PRs that are ready for review. For more info see the issue.

Closes #519

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,21 @@ on:
     - ready_for_review
 
 jobs:
+  build_single:
+    name: Build
+    if: github.event.pull_request.draft == true
+    runs-on: 'ubuntu-latest'
+    steps:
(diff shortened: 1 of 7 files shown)


stefsmeets:
@cwmeijer I updated the CI to always run linting / ubuntu with python 3.10 on all commits as we discussed.

cwmeijer (approved):
Looks great! No surprises. Makes a lot of sense. Thanks!

PR #536, opened 2023-03-28
add building for PRs in draft

cwmeijer:
I think this should turn on the build workflow for every commit to a PR whether it's in draft or not.
I know it's not the most eco-friendly but this way we get feedback much earlier which will help code development and prevent asking reviewers to review PRs with obvious errors.

_I see @stefsmeets is working to create the complete opposite of what I'm doing here. We should discuss this first :-)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ on:
     types:
     - opened
     - reopened
+    - edited
     - synchronize
     - ready_for_review
 


stefsmeets:
This function is currently not working (see #533) so all draft PRs are being build. 😅

My reasoning is that any PR in draft is still being worked on, so you can run tests on your local machine for testing. Only if a PR is ready for review do we want to run the CI.

Considering that our CI currently downloads roughly 30 GB of data (tensorflow + pytorch installs are huge), this emits something in the order of 5-10 kg of CO2 by my estimates for every single commit just for the data transfers. I think this is worth keeping in the back of our minds.

stefsmeets:
I'm closing this in favor of #533 as we discussed.

PR #537, opened 2023-03-28
Fix conflict with number of lines after imports between yapf and ruff

stefsmeets:
I noticed yapf and ruff have an incompatibility in the number of blank lines after imports. This PR configures yapf to add 2 lines after the imports so that it is compatible with ruff (and our previous isort config).


cwmeijer (approved):
Looks great. All checks pass. That's all you need for this PR I suppose. Sorry to have kept you waiting for this.

PR #538, opened 2023-03-28
Reduce install size of DIANNA

### LINKED ISSUE #510: Remove dependency on pytorch / tensorflow for default install of DIANNA
The default dianna install is quite large because of dependencies on tensorflow and pytorch. These dependencies are not necessary to run dianna. But it makes it very slow to install and the virtualenvironment is very large (6-7 GB).

It will improve the user experience if the default install of dianna is quite lean.

We could separate the dependencies where necessary in subpackages:

```
pip install dianna
pip install dianna[dashboard]
pip install dianna[tutorials]
...
```

stefsmeets:
In this PR I tried to reduce the install size of dianna. Turns out this is tricky because of hidden dependencies in tensorflow. There are a bunch of changes that I think are a step in the right direction, but I could not remove the dependency on tensorflow/pytorch unfortunately. For example, packages like `onnx_tf` have undeclared dependencies on tensorflow.

I sorted out the dependencies so that at least dianna can be installed with the least amount of packages. The dashboard and tutorials are now separate "extras" in the pip install (`pip install .[tutorials]` and `pip install .[dashboard]`.

My main goal is to make the CI a bit more efficient. It seems that this is mainly an issue on linux, which ships with a ton of gpu libraries (~2.8 GB) by [default](https://pytorch.org/get-started/locally/). On windows (~750 MB)/Mac (~850 MB) the install is much more efficient.

I switched out pytorch for the cpu version and tensorflow for tensorflow-cpu (thanks @WillemSpek for the suggestion!). On the CI this makes the installs significantly lighter. The linux cache is down to around 1 GB, and total run time is about 2 minutes. Way better than over 10 to 15 minutes we had a month ago! 🚀

This is also useful locally, to install dianna like so:
```
pip install torch --index-url https://download.pytorch.org/whl/cpu
pip install tensorflow-cpu
pip install -e .[dev]
```
On my linux install this helps reduce my venv size from ~5 GB (😬) to a managable 1 GB (still big, but what can you do 🤷‍♂️).

Closes #510

### Todo

- [x] Test whether cache is smaller (push empty commit)


cwmeijer:
Hey Stef,
Neither of those file ring a bell to me. I don't remember seeing them in our repo. We could ask the team today. I also see that this has only been a problem while checking in your last commit (which of course had nothing to do with that). Also, the main branch is passing fine. I'm so puzzled. I pressed 'rerun failed jobs' for now. Maybe it was some fluke.

stefsmeets:
My suspicion is that this fail happens somewhere in the other notebooks that are supposed to write a file. The file does not get written to the expected location, but it does not raise an error so pytest thinks it is OK.

The failing notebooks then depend on those files to exist, but they cannot find it. They have not been generated. I will have a look later today.

stefsmeets:
My bad, I accidentally removed `tf2onnx` as a dependency 😅
The notebooks spawn a subprocess to call `tf2onnx.convert`, so this does not actually trigger an exception that pytest can act on.

.github/actions/install-python-and-package/action.yml
```diff
@@ -16,20 +16,32 @@ runs:
...
-    
-    - name: Upgrade pip and install dependencies
+
+    - name: Upgrade pip
       shell: bash {0}
       run: |
         python3 -m pip install --upgrade pip setuptools wheel
+
+    # only necessary on linux to avoid bloated installs
+    - name: Install tensorflow/pytorch cpu version
```
cwmeijer, line 37: Great that we keep running gpu versions on Windows/Mac. This way our code is run with both cpu and gpu versions. Of course this can complicate things when certain builds start failing while others pass, but there's always something. Overall, I like it!

tests/utils.py
```diff
@@ -8,35 +8,118 @@
...
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 151, 254, 142, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 228, 254, 66, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 61, 251, 254, 66, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 141, 254, 205, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 215, 254, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 198, 176, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]],
```
cwmeijer, line 38: The formatting was specifically chosen like that, so you, as a human inspecting the testdata, can still easily see that we are looking at a "1". Could you revert the line wraps? If this is difficult, given the auto linters that are in place, we can think of something else, like saving the data as a csv or something. That's slightly harder to read, and will make for a slightly slower test, but would also still be fine.

tests/utils.py
```diff
@@ -68,7 +151,9 @@ def __init__(self, model_path, word_vector_file, max_filter_size):
         """
         self.filename = model_path
         # ensure the spacy english is downloaded
-        spacy.cli.download('en_core_web_sm')
+        model = 'en_core_web_sm'
+        if not spacy.util.is_package(model):
+            spacy.cli.download(model)
         self.tokenizer = SpacyTokenizer()
```
cwmeijer, line 157: Wow. Haha. That's a great! We should have done that from the start.

cwmeijer (changes requested):
Thanks a lot! Great improvements again!
I added a few comments. Only 1 is asking for an actual change though.

.github/actions/install-python-and-package/action.yml
stefsmeets: I don't know where it actually decides to run on the cpu or gpu. The VMs that the workflows run on do not have access to GPUs as far as I'm aware, so I think we are running cpu code only.

tests/utils.py
stefsmeets: Yeah, I was noticed that too. I will move the data to an ascii file and load it from there.

stefsmeets: I reformatted the table as plain text, would this be an acceptable solution for you?

cwmeijer (approved):
Haha great! Nice solution using the multiline string! Much nicer than it was before even.
I say merge it!

tests/utils.py
cwmeijer: absolutely!

PR #540, opened 2023-03-31
539 polish notebook rise timeseries

geek-yang:
In this PR, we re-organize the tutorial notebook of rise for timeseries to make it more user-friendly content-wise.

It will include two examples:
- Verify RISE for timeseries with a simple "expert" model
- Demonstrate RISE with a pretrained weather forecast (onnx) model


geek-yang:
Not sure why there are many commits named "
Merge branch 'main' of github.com:dianna-ai/dianna into main". I didn't do any merge operation in this PR. But I will squash all commits when merging to the main, to avoid this messy history.

geek-yang:
The second example in the notebook now is a binary forecast. I think it is nice to show a multi-variate case. We can change it when the multi-channel masking is available.

geek-yang:
Note for merging: Please use "squash and merge".

PR #548, opened 2023-04-06
Refactor dashboard

### LINKED ISSUE #418: Global cache in dashboard

### LINKED ISSUE #428: Stop DIANNA mid-way with a stop button
The dashboard now has a stop explanation button however, this only stops the dashboard loading, but does not affect the actual explanation run in DIANNA. The stop button should trigger DIANNA to stop its run. Does DIANNA have an inbuilt option to stop the run?

### LINKED ISSUE #511: Dashboard css file
Include a custom css file in the dashboard that we control and where we can consolidate the components we need to avoid style conflicts / inline style overrides

#504

### LINKED ISSUE #521: Dashboard: move update and stop button
Move the "update explanation" and "stop explanation" buttons to the location under the upload labels. See #454 for visual explanation

Related to #456 
Related to #479

### LINKED ISSUE #525: Dashboard: make individual method cards
Move the output of the different methods and their selection checkboxes into their own cards/containers.
See visual explanation in #454 

Related to #456 
Related to #479

### LINKED ISSUE #542: Make plan to change dashboard technology to increase testability and maintainability

stefsmeets:
This PR refactors the dashboard and moves it to streamlit.

Closes #542 

In addition:

Resolves #418 (caching is handled by st.cache)
Resolves #428 (stop button top right)
Resolves #511 (no longer necessary)
Resolves #521 (no longer relevant, dashboard auto-updates)
Resolves #525 (methods are in their own containers)

### Todo

- [x] Sort kernelshap issue
- [x] Organize model prepare step


stefsmeets:
Are you in the right branch? `app.run_server(debug=True)` is the way the old dashboard got started. There is no such code in this branch 😅

laurasootes:
never mind, I did not pull all changes yet.

laurasootes (approved):
Looks really good! The general outline already looks much nicer than the dash version. Also a lot less code, which makes it way more readable.

Some suggestions:
- I do get a whole lot of prints while running the dashboard for kernelshap, but can't find where it originates from (np.sum(w_aug), np.sum(self.kernelWeights) =, phi =). If you do know, can you remove it?
- Could you move all data to the data folder and remove the old app_data? (or the other way around)

I do have some aesthetic suggestions, but I think we can start from here and have another session with some people to decide how want the exact layout to look like.
One thing is that the colours for resnet don't look right yet and you can't see the original figure anymore, but I don't think that has to be fixed in this PR, since it is coming from the Dianna visualisation).

stefsmeets:
Thanks for the review. I checked the app data directory and I think that should all be OK. You may have existing data in the `app_data` directory, so that git did not remove it when you pulled the branch.

Let's create a new issue to follow up on the remaining points.

PR #554, opened 2023-04-13
514 add channel masking

cwmeijer:
Adds the ability to channel certain channels.

We can now 
- mask channels
- mask time steps
- make a combination of the above

We cannot do any completely random masking, which is fine I think, but could be discussed.
Note that I left some code duplication in the code for these different masking operations. I will change one of the 2 duplicates when I implement #546.

### DIFF
--- a/dianna/utils/maskers.py
+++ b/dianna/utils/maskers.py
@@ -4,7 +4,7 @@ import numpy as np
 
 
 def generate_masks(input_data: np.array, number_of_masks: int, p_keep: float = 0.5):
-    """Generate masks for time series data given a probability of keeping any time step unmasked.
+    """Generate masks for time series data given a probability of keeping any time step or channel unmasked.
 
     Args:
         input_data: Timeseries data to be explained.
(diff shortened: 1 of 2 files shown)


dianna/utils/maskers.py
```diff
@@ -14,9 +14,39 @@ def generate_masks(input_data: np.array, number_of_masks: int, p_keep: float = 0
...
     Single array containing all masks where the first dimension represents the batch.
     """
-    series_length = input_data.shape[0]
-    number_of_steps_masked = _determine_number_of_steps_masked(p_keep, series_length)
+    if input_data.shape[-1] == 1:  # univariate data
+        return generate_time_step_masks(input_data, number_of_masks, p_keep)
+
+    number_of_channel_masks = number_of_masks // 3
+    number_of_time_step_masks = number_of_channel_masks
+    number_of_combined_masks = number_of_masks - number_of_time_step_masks - number_of_channel_masks
```
geek-yang, line 22: This is quite an interesting way to implement the masking for multi-channels. I can understand that by doing this we can have a very balanced masking array, which contains masking for entire channels, masking for certain time steps across channels and a mixture of them. But I have a feeling that this makes it a bit too complex.

What I have in mind are two simple ways:
1. Simple flatten the input data and mask them brute-forcely
2. Loop through all channels and treat them individually (mask each channel separately and concatenate them)

I can imagine that the current implementation makes the segmentation very tricky to code. Let's have a chat about it. It could be possible that I misunderstand something.

But thanks a lot for the effort! This also provides more insight about what we want.

tests/methods/test_maskers.py
```diff
@@ -54,13 +55,76 @@ def test_mask_contains_correct_parts_are_mean_masked():
...
+
+def test_masking_univariate_leaves_anything_unmasked():
+    """Tests that something remains unmasked and some parts are masked for the univariate case."""
+    number_of_masks = 1
+    input_data = _get_univariate_input_data()
+
+    result = generate_masks(input_data, number_of_masks, 0.5)
+
+    assert np.any(result)
+    assert np.any(~result)
```
geek-yang, line 130: Smart check!

geek-yang (commented):
I left my comments below. Thanks for your effort @cwmeijer . I think we can have a discussion about it. I have some questions regarding the implementation. Just need to clarify a few things. Thanks in advance.

geek-yang (approved):
Thanks for the explanation @cwmeijer. To clarify a bit, this implementation tends to mask the timestep across all channels, a complete channel and a mixing of these two. This way we can have it more structured and it also allows us to implement the same strategy as we have for image masking based on wave function.

I checked the code and all the tests. They look very good to me. Nice work @cwmeijer! I think we can merge this now and continue working on the segmentation.

But since the algorithm is not so straight forward, we need to add more documentation/docstrings/examples to explain it. I will create a new issue for this and we also add this for the smart masking strategy based on wave functions for the images, which was not added before.

The purely random masking is a low-hanging fruit which can be another option to this. We can implement it later when working on Lime implementation for timeseries. I will create another issue for that.

Overall, nice work and thanks for all the strategical and smart thinking behind all the codes @cwmeijer 👍 .

PR #555, opened 2023-04-13
Add author to citation.cff and update readme

stefsmeets:
This PR adds myself as an author to the citation.ciff
I also fixed a small issue with the doi to zenodo, which was not pointing to the latest version.


cwmeijer (approved):
Nice!
Thanks for fixing the DOI. I thought we were already pointing to this 'general' DOI but maybe we only meant to do that and mistyped it anyway. I don't know. Great that it's fixed!

PR #562, opened 2023-04-18
546 masking time step segmentation

### LINKED ISSUE #546: smarter segmentation while creating masks in timeseries

cwmeijer:
Work in progress: Fixes #546.

This should now result in nice masks with clustered time steps.

time masks
(0=masked, 1=unmasked)
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 1 1 1 1 
^ channels/ time -->

(0=masked, 1=unmasked)
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 
^ channels/ time -->

(0=masked, 1=unmasked)
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 1 0 0 0 0 
^ channels/ time -->

```


Still needs to be done before merging:
- Refactoring / polishing
   - docstrings
   - default value for number_of_features should always be None (original series length). Clustering and projecting etc should be optional as it is probably not something the user would expect.

I spent most time in making the RISE test with the expert model pass. This now works for many cases. It can sometimes be a bit brittle for certain short series_lengths, for instance. In the general case it now seems to work reliably.

### CODE REVIEW

File: dianna/utils/projection.ipynb [hunk 1]
```diff
@@ -0,0 +1,179 @@
+{
```
geek-yang, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    masks = maskers._project_grids_to_masks(grids, masks_shape, offset=_offset)
</pre><p>Remove "v3".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dianna-ai/dianna/pull/562/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

dianna/methods/rise_timeseries.py
```diff
@@ -54,10 +55,13 @@ def explain(self,
             model_or_function, preprocess_function=self.preprocess_function)
         self.masks = generate_masks(input_timeseries,
                                     number_of_masks=self.n_masks,
+                                    feature_res=self.feature_res,
                                     p_keep=self.p_keep)
-        masked = mask_data(input_timeseries, self.masks, mask_type=mask_type)
+        self.masked = mask_data(input_timeseries,
```
geek-yang, line 60: Is it necessary to keep the masked data in the class? A copy of masked data could take some space since the input for timeseries is always the whole series (rather than one single instance in other cases).

dianna/utils/maskers.py
```diff
@@ -97,57 +96,185 @@ def _get_mask_value(data: np.array, mask_type: str) -> int:
...
+    masks_shape = (number_of_masks, *input_size)
+
+    if grid.shape == masks_shape:
+        masks = grid
+    else:
+        masks = _project_grids_to_masks(grid, masks_shape)
+    return masks.reshape(-1, *input_size, 1)
+
+
+def _project_grids_to_masks__old(grid: ndarray, masks_shape: tuple,
```
geek-yang, line 220: Nice to have both the old and new one, for comparison. The old function can be deleted afterward. (action point)

dianna/methods/rise_timeseries.py
cwmeijer: Good point! Storing it should be optional if at all. It can be convenient for a user to inspect the masks that were used, that's why it's in their in the first place.

dianna/utils/maskers.py
cwmeijer: deleted

dianna/methods/rise_timeseries.py
cwmeijer: implemented optional storing of these variables

dianna/utils/projection.ipynb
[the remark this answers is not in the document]
cwmeijer: I prefer not to make a test out of it right now. It's not testing anything right now, just visualizing, it would a bit hard to come up with a test that is not just duplicating the code under test. The code is also a bit deep, in the sense that it is an implementation detail. The code in ```_project_grids_to_masks``` can very well be changed or replaced by something quicker. By then, this code itself can serve as a unit test.

For now, I removed the notebook entirely, which was indeed only meant for fiddling while debugging.

cwmeijer:
For anyone reviewing this PR. Please have a look first at my blog explaining this approach on a high level: https://medium.com/@cwmeijer/how-to-correctly-mask-time-series-data-to-use-with-xai-90247ac252b4

dianna/utils/maskers.py
```diff
@@ -1,45 +1,58 @@
...
+def generate_masks(
+    input_data: np.array,
+    number_of_masks: int,
+    feature_res: int = 8,
+    p_keep: float = 0.5,
+):
     """Generate masks for time series data given a probability of keeping any time step or channel unmasked.
 
     Args:
         input_data: Timeseries data to be explained.
```
SarahAlidoost, line 18:
```suggestion
        input_data: Timeseries data to be masked.
```

SarahAlidoost: I see that the shape information is obtained by hardcoding the index in many places in the code for example  `number_of_channels = input_data.shape[1]`. Shall we add a text in the docstring what is the expected shape of `input_data`?

dianna/utils/maskers.py
```diff
@@ -97,57 +96,179 @@ def _get_mask_value(data: np.array, mask_type: str) -> int:
...
+        p_keep, len(time_indices))
+    top_indices = heapq.nsmallest(number_of_unmasked_cells,
+                                  time_indices,
+                                  key=lambda time_step: flat[time_step])
+    flat_mask = np.ones(flat.shape, dtype=bool)
+    flat_mask[top_indices] = False
+    return flat_mask.reshape(float_mask.shape)
+
+
+def _generate_interpolated_float_masks_for_image(input_size: int,
```
SarahAlidoost, line 164: I couldn't understand the reason why this function is a private function with an underscore in its name. I see that it is used in `rise_image` module. Can we make the name without underscore?

dianna/utils/maskers.py
```diff
@@ -1,45 +1,58 @@
...
     Single array containing all masks where the first dimension represents the batch.
     """
     if input_data.shape[-1] == 1:  # univariate data
-        return generate_time_step_masks(input_data, number_of_masks, p_keep)
+        return generate_time_step_masks(input_data,
+                                        number_of_masks,
+                                        p_keep,
+                                        number_of_features=feature_res)
 
     number_of_channel_masks = number_of_masks // 3
```
SarahAlidoost, line 32: can you please add an inline comment and explain `3` in `number_of_masks // 3`?

tests/methods/test_maskers.py
```diff
@@ -1,7 +1,9 @@
 import numpy as np
 import pytest
+from pandas import DataFrame
 from dianna.utils.maskers import generate_channel_masks
 from dianna.utils.maskers import generate_masks
+from dianna.utils.maskers import generate_time_step_masks
```
SarahAlidoost, line 6: can we also test the function `_generate_interpolated_float_masks_for_image`?

tests/methods/test_maskers.py
```diff
@@ -10,37 +12,78 @@ def test_mask_has_correct_shape_univariate():
...
+    number_of_masks = 5
+
+    result = generate_masks(input_data, number_of_masks=number_of_masks)
+
+    assert result.dtype == bool
+
+
+def test_generate_time_step_masks_dtype_multivariate():
+    """Test masked data has the correct dtype for a multivariate input."""
+    input_data = _get_multivariate_input_data()
```
SarahAlidoost, line 33: I couldn't understand what is meant by `multivariate_input_data`. Does it mean 2D data or more data with more than two dimensions? Currently, the function `_get_multivariate_input_data` only generates 2D data. When I changed it to 3D data, the tests failed.

tests/methods/test_maskers.py
```diff
@@ -10,37 +12,78 @@ def test_mask_has_correct_shape_univariate():
     input_data = _get_univariate_input_data()
     number_of_masks = 5
 
-    result = _call_masking_function(input_data, number_of_masks=number_of_masks)
+    result = _call_masking_function(input_data,
+                                    number_of_masks=number_of_masks)
```
SarahAlidoost, line 16: if the test only checks the shape, calling `generate_masks(input_data, number_of_masks=number_of_masks)` is enough instead of `_call_masking_function`

SarahAlidoost (commented):
@cwmeijer thank you for implementing this. I read the blog and checked the implementation. The code can benefit from more documentation, but I saw that this will be addressed in a different issue later. I left some minor suggestions/comments.

dianna/utils/maskers.py
cwmeijer: Good suggestion. I changed the function to public (removed the underscore).

dianna/utils/maskers.py
cwmeijer: I added a comment.

tests/methods/test_maskers.py
cwmeijer: I meant multivariate time series. That is still 2D because we have time steps * channels. 
I changed the function title from ```_get_multivariate_input_data``` to ```_get_multivariate_time_series``` to make it clear that it only gets time series data. 
Let me know if you think I should clarify this some more.

PR #563, opened 2023-04-18
Fix failing workflows

stefsmeets:
I noticed two workflows are failing on the main branch:

```
Invalid workflow file: .github/workflows/fair-software.yml#L33
The workflow is not valid. .github/workflows/fair-software.yml (Line: 33, Col: 9): Unexpected symbol: 'and'. Located at position 39 within expression: (github.event_name == 'pull_request') and (github.event.pull_request.draft == false)
```

```
Invalid workflow file: .github/workflows/documentation.yml#L28
The workflow is not valid. .github/workflows/documentation.yml (Line: 28, Col: 9): Unexpected value 'extras-require'
```

This PR fixes these.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -25,7 +25,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/install-python-and-package
-        extras-require: docs
+        with:
+          extras-require: docs
       - name: Install pandoc using apt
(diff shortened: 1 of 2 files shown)


cwmeijer (approved):
I quickly looked up the syntax corrections. This looks conform the manual now (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepswith). Great job! Thanks!

PR #576, opened 2023-04-25
574 README taxonomy

WillemSpek:
Updated the taxonomy according to the recent advancement for time-series. Additionally, DIANNA does not (yet) have global explanations, therefore these are now excluded in the taxonomy. 

I used the figure from here: https://www.semanticscholar.org/paper/Classification-of-Explainable-Artificial-Methods-Vilone-Longo/47270bad365ad9c4c4b13c9f56fa384e5160c072 Given that we already used this image, I assumed there were no copyright issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ DIANNA software is addressing needs of both (X)AI researchers and mostly the var
 
 After studying the vast XAI landscape we have made choices in the parts of the [XAI Taxonomy](https://doi.org/10.3390/make3030032) on which methods, data modalities and problems types to focus. Our choices, based on the largest usage in scientific literature, are shown graphically in the XAI taxonomy below:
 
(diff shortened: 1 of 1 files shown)


cwmeijer (approved):
Thanks! Much better!
I think our output type actually is both visual and numerical. The latter is still missing in the figure but that can also be fixed some other time.
Merge it :-)

geek-yang (commented):
The markdown check is failing because one dead link. It doesn't relate to this PR but maybe it is better to fix it here.

The dead link is related to pytorch lightning, it should be "https://lightning.ai/docs/pytorch/latest/api/lightning.pytorch.core.LightningModule.html#lightning.pytorch.core.LightningModule" now.

@WillemSpek sorry for the small hassle 😜.

WillemSpek:
Thanks for the linter fix @geek-yang! Will merge now

PR #578, opened 2023-05-01
557 fix first instance issue for Lime timeseries

### LINKED ISSUE #557: Fix first instance issue in LIME timeseries implementation
`Lime` explainer (`lime_base`) requires that the first instance in masked input should be the original data. Check this link for more information https://github.com/marcotcr/lime/blob/fd7eb2e6f760619c29fca0187c07b82157601b32/lime/lime_base.py#L148

As a result, we should modify our code to make sure that the first instance in `masked`, `predictions` and `distance` are corresponding to the original input. (e.g. first instance of `distance` must all be 1).

geek-yang:
Fix #557.

Now it ensures the first instance is the original data, which satisfies the requirement from https://github.com/marcotcr/lime/blob/fd7eb2e6f760619c29fca0187c07b82157601b32/lime/lime_base.py#L148-L149

### DIFF
--- a/dianna/methods/lime_timeseries.py
+++ b/dianna/methods/lime_timeseries.py
@@ -83,6 +83,10 @@ class LIMETimeseries:
             model_or_function, preprocess_function=self.preprocess_function
         )
         masks = generate_masks(input_timeseries, num_samples, p_keep=0.9)
+        # NOTE: Required by `lime_base` explainer since the first instance must be the original data
+        # For more details, check this link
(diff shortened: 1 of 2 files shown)


cpranav93 (approved):
It looks great! One point of modularity, it might be better to have the "masks[0,:,:] = 1.0" as a flagged option in maskers.py rather than in lime. But we can also do that later once Chris is finished with the masks.

PR #581, opened 2023-05-03
Test user workflow in dashboard

### LINKED ISSUE #567: Tests for the dashboard
It would be useful if the dashboard could be tested.

See: https://blog.streamlit.io/testing-streamlit-apps-using-seleniumbase/

stefsmeets:
This PR adds a test module that tests the user workflow in the dashboard using [playwright](https://playwright.dev).

These tests make sure that the dashboard runs given the example data. It makes sure that the workflow completes (i.e. the RISE method produces an explanation) by checking whether the image or text is there.

These tests are somewhat time consuming, so I run them in a separate github action on python3.10/ubuntu.

Closes #567

### Todo

- [x] Add playwright to dependencies
- [x] Add pytest marker for dashboard
- [x] Add github action workflow (dashboard marker only)
- [x] Skip dashboard by default: https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option


stefsmeets:
Hi @cpranav93 @laurasootes Would either of you be willing to review this PR?

Don't mind the sonarcloud test failing, I changed 3 lines in cli.py and it complains that the test coverage on these is 0% 😬.

laurasootes:
Hmm if I try to run `pytest -v -m dashboard`, the tests fail because of a connection refused error (both for LOCAL True/False). I am not sure why 

I have 2 questions:
- why do you only test on ubuntu?
- where are the browsers defined that are used for the tests?

stefsmeets:
Which OS are you running this on? I only tested it under linux. I can have a look whether it works on Windows.

My idea was that you can set `LOCAL=True` if you start the dashboard yourself in a separate tab. I found this convenient to be able to iterate a bit faster while developing this. Otherwise, the tests will start the dashboard in a subprocess, which is a bit slower.

1. I test in Ubuntu because it's the most efficient runner on Github. The goal of the PR is to test the user workflow in the browser. I don't think we gain anything by testing in another OS. The functionality/correctness is verified via the other tests.
2. The browser can be defined via the `--browser` switch in the command line. The default is `chromium`, so it picks that.

stefsmeets:
I guess these tests do not necessarily work 'out-of-the-box'. They need some fairly substantial setup, before you can use them (i.e. install the browser). 

Do you think we should disable them by default, and just let github actions run them?

laurasootes:
I am using Mac M1, and was not able to install chromium within a few minutes :unamused:. I think it would indeed be better to only include the test in the GitHub actions.

I agree with your reasoning on only testing on ubuntu, but would suggest to also include webkit and firefox for the browsers, just to be sure.

laurasootes:
image

It appears that the dashboard tests are still skipped?

stefsmeets:
Thanks for pointing that out, it's surprisingly tricky to set this up. I added a flag to tell pytest to also run the dashboard tests:

```
pytest --dashboard
```

If this flag is not specified, it will mark all dashboard tests as skip. I hope this works now.

laurasootes:
Yes it works now 🥳 
It runs all with the --dashboard flag, and skips the dashboard tests without it

Sounds good to only do chromium btw

Should this flag option maybe be added to the dev documentation?

laurasootes (approved):

stefsmeets:
Cheers, I will add a small section to the documentation.

PR #584, opened 2023-05-09
Notebooks

APJansen:
I've made all the notebooks in the tutorials folder runnable from colab, by adding at the start of each of them the code block below, only varying the `paths_to_download` between them. I did this rather than cloning the entire repo as the latter took the setup time from almost 1 to almost 2 minutes.

In lime_images, the example image uses a variable `true_species` suggesting other options, but in the data there is only the one given, so I hardcoded the colab setup to download that one.

I've tested that they all run on colab. Having some issues running some locally, mostly complaints that lime doesn't exist, but that must be something outdated in my environment, as nothing changed when it runs locally.

Also added links on the tutorial page.

```python
running_in_colab = 'google.colab' in str(get_ipython())
if running_in_colab:
  # install dianna
  !python3 -m pip install dianna
  
  # download data used in this demo
  import os 
  base_url = 'https://raw.githubusercontent.com/dianna-ai/dianna/main/tutorials/'
  paths_to_download = ['models/season_prediction_model_temp_max_binary.onnx']
  for path in paths_to_download:
      local_directory = os.path.dirname(path)
      os.makedirs(local_directory, exist_ok=True)
      !wget {base_url + path} -O {path}
```


APJansen:
Btw, the colab buttons point to the notebooks in the main branch, so before this is merged they won't work.

PR #585, opened 2023-05-09
Fix the special character problem

stefsmeets:
This PR attempts to fix the 'special character problem'.

I have identified this as a failing string:

`Hello, to the world!`

Oddly enough, this works:

`Hello, world!`

I updated the fix in `SpacyTokenizer` to fix the sentence before sending it to the spacy tokenizer instead of trying to make sence of the tokens after tokenization. The new fix will add spaces around all instances of UNKWORDZ where missing using regexes. And hopefully, it's a bit more future-proof when we come accros new issues.

The implementation differs slightly in that this does not consider multiple special characters as a single token. This is inline with the tokenization expected from spacy.

I found the error message very confusing, so I also added a check to make sure that the number of tokens is consistent for all sentences.

I also updated the tests to specifically test the tokenizer, rather than the full explain text pipeline.

## Todo

- [x] Adjust tokenizer in dashboard
- [x] Use compiled regexes

image


stefsmeets:
Hi @laurasootes Can you have a look at this PR and let me know what you think?

The Ubuntu/Python3.7 build is failing because of #587

PR #591, opened 2023-05-12
569 multivariate timeseries visualization

cwmeijer:
Adds multivariate support to the timeseries visualization:

image

### DIFF
--- a/dianna/visualization/timeseries.py
+++ b/dianna/visualization/timeseries.py
@@ -1,7 +1,9 @@
 from typing import Any
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import Union
 import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import cm
@@ -11,8 +13,8 @@ def plot_timeseries(
     x: np.ndarray,
     y: np.ndarray,
     segments: List[Dict[str, Any]],
-    xlabel: str = 'x',
-    ylabel: str = 'y',
+    x_label: str = 'x',
+    y_label: Union[str, Iterable[str]] = None,
     cmap: Optional[str] = None,
     show_plot: bool = False,
     output_filename: Optional[str] = None,
@@ -28,8 +30,8 @@ def plot_timeseries(
             'weight'. Here, `index` is the index of the segment of feature,
             `start` and `end` determine the location of the
             segment, and `weight` determines the color.
-        xlabel (str, optional): Label for the x-axis
-        ylabel (str, optional): Label for the y-axis
+        x_label (str, optional): Label for the x-axis
+        y_label (str, optional): Label for the y-axis
         cmap (str, optional): Matplotlib colormap
         show_plot (bool, optional): Shows plot if true (for testing or writing
             plots to disk instead).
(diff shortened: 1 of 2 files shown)


elboyran:
@loostrum do you have a chance to review this?

dianna/visualization/timeseries.py
```diff
@@ -11,13 +13,12 @@ def plot_timeseries(
...
+    y_label: Union[str, Iterable[str]] = None,
     cmap: Optional[str] = None,
     show_plot: bool = False,
     output_filename: Optional[str] = None,
-    ax: Optional[plt.Axes] = None,
-) -> plt.Axes:
+) -> plt.Figure:
     """Plot timeseries with segments highlighted.
 
     Args:
```
stefsmeets, line 24: I think it is worth spelling out the expected shapes for the monovariabe/multivariate data here. This function expects channels to be on the first dimension. This may not be obvious to someone used to working with columnar data like myself, who will expect channels to be on the 2nd dimension.

stefsmeets (changes requested):
I'm not entirely happy with this PR.

1. The docstring for `plot_timeseries` does not entirely cover the functionality. Could you update it to make it more clear how the function works with multivariate data?

2. The interface is not the same as before for monovariate data. To me it was not directly obvious what changes I needed to make to get it to work. I had to fiddle with x and y, and adjust the segments to get it to work. Better documentation and checks to make it clear what is wrong would help.

That said, I'm thinking that it may be worth to consider splitting plot_timeseries into `plot_timeseries` and `plot_timeseries_multi` to keep the API clear. To me, this function is not intuitive at the moment. Part of the reason is also having to define segments each time to call it, this could be integrated into the `explain_timeseries` function. Not to be addressed in this PR, but definitely something to follow up on.

4. Can you also have another look at the notebooks? The notebooks break whenever the `plot_timeseries` function is called on my end:

`ValueError: x and y must have same first dimension, but have shapes (28,) and (1,)`

I think either the segments are not correctly defined or the shape of the input data is not right.

tests/test_visualization.py
```diff
@@ -1,18 +1,64 @@
...
+    x = np.linspace(start=0, stop=10, num=20)
+    y = np.stack((np.sin(x), np.cos(x), np.tan(0.4 * x)))
+    segments = get_test_segments(data=y)
+    output_path = Path(tmpdir) / 'temp_visualization_test_multivariate.png'
+
+    plot_timeseries(x=x,
+                    y=y,
+                    segments=segments,
+                    show_plot=False,
+                    output_filename=output_path)
```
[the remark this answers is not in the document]
cwmeijer: Updated the variable name. Left the Path() /  construction because I like it better although it's almost the same.

dianna/visualization/timeseries.py
```diff
@@ -28,46 +29,73 @@ def plot_timeseries(
             'weight'. Here, `index` is the index of the segment of feature,
```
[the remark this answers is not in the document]
cwmeijer: included

dianna/visualization/timeseries.py
cwmeijer: I added it in the docstring for y

tests/test_common_usage.py
```diff
@@ -6,7 +6,7 @@
 
 def test_common_RISE_image_pipeline():  # noqa: N802 ignore case
     """No errors thrown while creating a relevance map and visualizing it."""
-    input_image = np.random.random((5, 5, 3))
+    input_image = np.random.random((224, 224, 3))
```
stefsmeets, line 9: This would be a regression from https://github.com/dianna-ai/dianna/pull/606 which we merged yesterday :-)

```suggestion
    input_image = np.random.random((5, 5, 3))
```

stefsmeets (approved):
Nice work, looks good to me!

PR #606, opened 2023-05-30
Improve test suite time by 25%

stefsmeets:
Hey, I noticed that some of the tests are quite slow. I had a quick look and noticed the RISE image pipeline test can be made about 20 times faster by reducing the shape of the image:

Before:
`28.10s call     tests/test_common_usage.py::test_common_RISE_image_pipeline`

After:
`1.74s call     tests/test_common_usage.py::test_common_RISE_image_pipeline`

Since we are testing the pipeline only, the size of the image is irrelevant (it's filled with random data anyway). This takes the time needed to run the tests down from 0:01:59 to 0:01:31 on my setup.

### DIFF
--- a/tests/test_common_usage.py
+++ b/tests/test_common_usage.py
@@ -6,7 +6,7 @@ from tests.utils import run_model
 
 def test_common_RISE_image_pipeline():  # noqa: N802 ignore case
     """No errors thrown while creating a relevance map and visualizing it."""
-    input_image = np.random.random((224, 224, 3))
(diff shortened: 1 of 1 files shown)


geek-yang (approved):
Well spot! Nice change. I simply time the execution with `time pytest tests/test_common_usage.py::test_common_RISE_image_pipeline` and the actual time for the process is "0.852s" and "0.158s", before and after. 😎 Good to have this improved!

Issue #613, opened 2023-06-01
Compatibility issues 


WillemSpek:
I have some compatibility issues with scikit-image 0.19.1, as Quantus requires 0.19.3 or higher. I think the current scikit-image requirement of ==0.19.1 is too stringent and needs to be relaxed. I just did a quick grep and saw we only use it in rise.py

`willem@debian:~/Documents/dianna/dianna$ grep -R --exclude-dir=lib skimage *`
`bin/skivi:from skimage.scripts.skivi import main`
`methods/kernelshap.py:import skimage.segmentation`
`methods/kernelshap.py:        Check keyword arguments for the skimage.segmentation.slic function`
`methods/kernelshap.py:        https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.slic`
`methods/kernelshap.py:        image_segments = skimage.segmentation.slic(image=image,`
`methods/kernelshap.py:        https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.slic`
`methods/kernelshap.py:            skimage.segmentation.slic, kwargs)`
`methods/rise.py:from skimage.transform import resize`
`grep: methods/__pycache__/kernelshap.cpython-39.pyc: binary file matches`
`grep: methods/__pycache__/rise.cpython-39.pyc: binary file matches`

Aditionally I have a naming issue in kernelshap.py where `model` is used instead of `model_or_function`. For the sake of consistency and naming collisions I would like to change this one as well to `model`.

PR #614, opened 2023-06-01
upped scikit-image and some naming changes

WillemSpek:
see #613 it's just some very minor changes that should enhance compatibility and not impede functionality. Should be a quick review :)

### DIFF
--- a/dianna/methods/kernelshap.py
+++ b/dianna/methods/kernelshap.py
@@ -54,7 +54,7 @@ class KERNELSHAPImage:
 
     def explain(
         self,
-        model,
+        model_or_function,
         input_data,
         labels,
         nsamples='auto',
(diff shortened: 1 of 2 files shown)


stefsmeets (approved):
What exactly does this fix?

WillemSpek:
Thanks for the review! 
1. Dependency coonflict

`ERROR: Cannot install -r requirements.in (line 3) and -r requirements.in (line 7) because these package versions have conflicting dependencies.`

`The conflict is caused by:`
    `dianna 0.9.0 depends on scikit-image==0.19.1`
    `quantus 0.4.0 depends on scikit-image>=0.19.3`
    
2. problem with Quantus and DIANNA integration:

`TypeError: explain() got multiple values for argument 'model'`

Basically because quantus also uses model to pass on to the explain function which collides with the required positional argument model from DIANNA in kernelshap.explain

stefsmeets:
Thanks for the explanation, makes sense :+1:

PR #588, opened 2023-05-10
Drop support for python 3.7

stefsmeets:
This PR drops support for Python 3.7.

Closes #587 

Cool new python features we get from supporting [3.8](https://www.python.org/downloads/release/python-380/) and up:

- Walrus operator
- `=` specifier in f-strings

PR #615, opened 2023-06-05
Fix readthedocs build

stefsmeets:
This PR fixes the rtd build by bumping the python version to 3.9.

This was broken, because we dropped support for Python 3.7 recently (#588), and rtd by default builds in 3.7.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,4 +1,9 @@
 version: 2
+
+build:
+  tools:
+    python: "3.9"
+
 python:
   install:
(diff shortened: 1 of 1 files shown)


elboyran:
[like]  Elena Ranguelova reacted to your message:
________________________________

elboyran:
[0]     Elena Ranguelova reacted to your message:
________________________________

elboyran:
@stefsmeets I see that you fix the build yesterday, but still some checks on main are not passing.

PR #616, opened 2023-06-06
Exclude coverage check for dashboard

geek-yang:
`dashboard` is checked but since it is checked within subprocess it is not detected by sonarcloud coverage. This produces misleading results. So it is better to exclude the coverage check for dashboard from the report.


stefsmeets (approved):
Super!

Issue #620, opened 2023-06-13
IME wrapper gets rid of coefficients and just outputs masks


WillemSpek:
Whilst scouring through the LIME implementation both by in the `lime` package and the wrapper built in `dianna`, I found out that dianna does not return the original feature importances but rather masks that indicate whether the feature contributes positively, negatively or not at all. As a result, explanation quality suffers severely, as I will show below. Luckily, there is an easy fix by simply grabbing the `ImageExplanation` object and computing the original feature values from there and putting this in the segments, I will write a method in our wrapper that does this for us. 

Running LIME without the correction:
![bee_not_corrected](https://github.com/dianna-ai/dianna/assets/33344129/f162f775-4544-422d-b88a-25727f734056)

Running LIME with the proposed correction:
![bee_corrected](https://github.com/dianna-ai/dianna/assets/33344129/68b843f7-003d-45c7-84e7-4c5a6af676bf)

PR #621, opened 2023-06-13
solved LIME coefficients

WillemSpek:
See issue #620 the current LIME implementation does not make use of the coefficients that LIME actually yields and instead returns the explanation as a mask with values -1, 0 and 1 that designates between negative, neutral and positive contribution. The implementation for this masking also seems a bit slow and could be more vectorized. I propose my change that grabs the original coefficients from the surrogate model in LIME that also yield for much more granular explanations as I will show below. 

`quickshift`:
With masks:
bee_not_corrected

`skimage.segmentation.slic`: 
With masks:
bee_masked_kmeans

With LIME surrogate model coeffficients:
bee_corrected


geek-yang (commented):
@WillemSpek Just a very quick comment before diving into the code. You also need to write a unit test for the `get_explanation_values` function, otherwise the coverage check will fail.

I don't have much time today and tomorrow. If nobody leaves any review, I can take a look on Thursday.

Thanks a lot for noticing this, and contributing to the code 😃.

cwmeijer:
In addition to Yang's comments. The results look so much better!! Great job!
Let us know when you added some test(s). So we can review it.

WillemSpek:
Thanks for all the input, though a big culrpit of the poor result is apparently the segmenation algorithm: `quickshift` is used by LIME by default and the image I used uses `skimage.segmentation.slic` which is k-means centroid clustering. Nevertheless, the expalantion still looks better with the full coefficients imho, I have updated the PR description accordingly :)

WillemSpek:
Sorry for all the redundant commits, was struggling a bit with the testing and the random nature of things. All should be good now :)

I have added a simple unittest `test_lime_values` in `tests/test_lime.py` similar to `test_lime_function.py`.

dianna/methods/lime.py
```diff
@@ -1,6 +1,8 @@
 import numpy as np
+from lime.lime_image import ImageExplanation
 from lime.lime_image import LimeImageExplainer
 from lime.lime_text import LimeTextExplainer
+from numpy.typing import NDArray
```
geek-yang, line 5:
```suggestion
```
Not necessary. The returned type can just be `np.array`, see below.

dianna/methods/lime.py
```diff
@@ -276,3 +282,27 @@ def moveaxis_function(data):
         if self.preprocess_function is None:
             return moveaxis_function
         return lambda data: self.preprocess_function(moveaxis_function(data))
+
+    @staticmethod
```
geek-yang, line 286:
```suggestion
```
A hidden function would be enough since we encourage users to call dianna API rather than call the explainer directly, so not necessary for a static method.

dianna/methods/lime.py
```diff
@@ -276,3 +282,27 @@ def moveaxis_function(data):
...
+    def get_explanation_values(label: int, explanation: ImageExplanation) -> NDArray:
```
geek-yang, line 287:
```suggestion
    def _get_explanation_values(self, label: int, explanation: ImageExplanation) -> np.array:
```
Make it a hidden function.

dianna/methods/lime.py
```diff
@@ -205,12 +209,14 @@ def explain(self,
...
-                                                num_features=num_features, **get_image_and_mask_kwargs)[1]
-                 for label in labels]
-        return masks
+        if return_masks:
+            get_image_and_mask_kwargs = utils.get_kwargs_applicable_to_function(explanation.get_image_and_mask, kwargs)
+            maps = [explanation.get_image_and_mask(label, positive_only=positive_only, hide_rest=hide_rest,
+                                                   num_features=num_features, **get_image_and_mask_kwargs)[1]
+                    for label in labels]
+        else:
+            maps = [self.get_explanation_values(label, explanation) for label in labels]
```
geek-yang, line 218:
```suggestion
            maps = [self._get_explanation_values(label, explanation) for label in labels]
```

dianna/methods/lime.py
```diff
@@ -179,6 +182,7 @@ def explain(self,
             top_labels: Top labels
             num_features (int): Number of features
             num_samples (int): Number of samples
+            return_masks (bool): Return discretized masks or LIME values
```
geek-yang, line 185:
```suggestion
            return_masks (bool): If true, return discretized masks. Otherwise, return LIME scores
```

dianna/methods/lime.py
```diff
@@ -276,3 +282,27 @@ def moveaxis_function(data):
         if self.preprocess_function is None:
             return moveaxis_function
         return lambda data: self.preprocess_function(moveaxis_function(data))
+
+    @staticmethod
+    def get_explanation_values(label: int, explanation: ImageExplanation) -> NDArray:
+        """Get the values from LIME in a salience map.
```
geek-yang, line 288:
```suggestion
        """Get the importance scores from LIME in a salience map.
```

dianna/methods/lime.py
```diff
@@ -276,3 +282,27 @@ def moveaxis_function(data):
...
+
+        Leverages the `ImageExplanation` class from LIME to generate salience maps
+        based on the segmentation that was produced on the images, also stored in
+        `ImageExplanation`.
```
geek-yang, line 292: Please also consider adding more explanation about the scores, e.g. "The importance values are weight of surrogate models (e.g. weight of a ridge model) which are used by LIME to evaluate the contribution of each data/pixel/segmentation to the predicted classes."

geek-yang (changes requested):
Good job @WillemSpek ! Sorry for the delay. I have tested your implementation and it works well on my machine. It is good to give the user option to grab the importance scores from LIME.

Btw, I won't put a "bug" label on this PR since it makes sense to only show user the masks. For instance, most of the example notebooks from LIME also play with the segmentation masks only, rather than including importance scores (e.g. https://github.com/marcotcr/lime/blob/master/doc/notebooks/Tutorial%20-%20images%20-%20Pytorch.ipynb). 

There is some logic behind it. LIME uses linear surrogate model to delegate the local importance of each data point/pixel to the predicted classes. To train these surrogate models, we need to generate many sample data around each data point. It is very tricky to define the neighborhood and perturb your data to get good samples. Here are the explanations about how to define a meaningful neighborhood https://christophm.github.io/interpretable-ml-book/lime.html and how to make a good explainer https://slds-lmu.github.io/iml_methods_limitations/lime.html.

To conclude, the importance scores, which are basically the weight of your surrogate models, are very sensitive to the perturbations. Therefore, the scores are not robust and simply displaying these scores could be sometimes very misleading, especially to those who have little knowledge about LIME. Therefore, even in the documentation/tutorials from the original LIME implementation, they always show segmentation masks rather than scores.

But I agree with you that for some advanced users, it is necessary to have these scores. Thanks a lot for the contribution. Please address all my comments (simply by adding commits should be fine) and test it again before merging it to the code base. I will give you a go afterwards.

geek-yang:
Btw, I will remove `bug` tag from this PR and the issue. But thanks for your thought 👍.

WillemSpek:
Thanks so much for the review @geek-yang! Your comment on the coefficient stability is also very insightful for me and I agree that from a user perspective, only returning the masks makes more sense. My usecase (metrics) does require coefficient scores for instance as discretizing the output leads to some loss of granularity which affects the metrics. Thanks for all the review comments, I will dive into them!

dianna/methods/lime.py
WillemSpek: Ah yes, I have used this from the np.typing module as it supports hinting towards dimensions and types, but I never end up doing that. np.array is a higher level function to create an np.ndarray object which is what all arrays in numpy are defined as. So shouldn't we use `np.ndarray`?

geek-yang: Yes, I think `np.ndarray` is much better. Thanks for the suggestion.

geek-yang (approved):
Nice work! Thanks for addressing my comments. Looks good now!

dianna/methods/lime.py
WillemSpek: I guess this is now consistent for now with the rest of the codebase so lets leave it as `np.array` for now but maybe this could be listed as a more general issue.

PR #622, opened 2023-07-11
refactor masker for RISEImage into free function

egpbos:
This is necessary for distance_explainer to be able to reuse it.


egpbos:
Did force push to fix linter issues.

PR #628, opened 2023-07-25
Update README.md

elboyran:
Fixed links to TS tutorials


elboyran:
Yes, I had the same problem and coudn't fix it :-( That's why I was hooping someone else can spot the problem.

geek-yang (commented):
> Yes, I had the same problem and coudn't fix it :-( That's why I was hooping someone else can spot the problem.

I suspect that this is due to the server side rendering trick of github. Currently the image is requested via `[Weather Logo]`, which fetches image from assets and ends up with rendering, with the link to the notebook unresolved.

To fix this, I guess we need to use the raw link to the image, rather than via assets (might be relevant with webpack thingy? apologize for my limited web knowledge 😓), for instance `src="https://user-images.githubusercontent.com/3244249/242001499-3ff3d639-ed2f-4a38-b7ac-957c984bce9f.png"`

geek-yang:
@elboyran Sorry for pushing commits to your PR. Now it seems to work!

cwmeijer (approved):
It works now! 🏆

Let's get it merged!

PR #632, opened 2023-08-07
fixed number of segments bug

### LINKED ISSUE #631: KernelSHAP doesn't yield the correct amount of Shapeley values
<h2> Problem description </h2>

This is a bug I encountered during my research. The skimage.segementation.slic method doesn't guarantee that the amount of segments is equal to the `n_segments` that are provided. As a result, the way `kernelshap` is set up it sometimes misses shapeley values. This is hinted towards in [their docs](https://scikit-image.org/docs/stable/api/skimage.segmentation.html#skimage.segmentation.slic) 

`>>> from skimage.segmentation import slic`
 `>>> segments = slic(X_test, n_segments=90)`
 `>>> np.unique(segments).size `
 `>>> 88`
 
 

<h2> Proposed solution </h2>

Instead of using the `n_segments` parameter, we simply count the amount of clusters after creating them as I have done in the problem description. We'll use this to pass onto the kernelshap explainer.

WillemSpek:
This PR fixes #631. In some cases, kernelshap would over- or (worse) underallocate memory for the shapeley values leading to issues. This PR should fix this bug.

### DIFF
--- a/dianna/methods/kernelshap.py
+++ b/dianna/methods/kernelshap.py
@@ -116,6 +116,7 @@ class KERNELSHAPImage:
         self.image_segments = self._segment_image(self.input_data, n_segments,
                                                   compactness, sigma,
                                                   **slic_kwargs)
+        n_segments = np.unique(self.image_segments).size
 
         # call the Kernel SHAP explainer
         explainer = shap.KernelExplainer(
--- a/tests/test_kernelshap.py
+++ b/tests/test_kernelshap.py
@@ -1,5 +1,6 @@
 from unittest import TestCase
 import numpy as np
+import skimage.segmentation
 from dianna.methods.kernelshap import KERNELSHAPImage
 
 
@@ -63,5 +64,7 @@ class ShapOnImages(TestCase):
             compactness=10.0,
             sigma=0,
         )
+        segments = skimage.segmentation.slic(input_data, n_segments=n_segments)
+        n_segments = np.unique(segments).size
 
         assert shap_values[0].shape == np.zeros((1, n_segments)).shape
--- a/tests/test_kernelshap.py
+++ b/tests/test_kernelshap.py
@@ -64,7 +64,11 @@ class ShapOnImages(TestCase):
             compactness=10.0,
             sigma=0,
         )
-        segments = skimage.segmentation.slic(input_data, n_segments=n_segments)
+        segments = skimage.segmentation.slic(image=input_data.reshape(28, 28, 1),
+                                             n_segments=n_segments,
+                                             compactness=10.0,
+                                             sigma=0.)
         n_segments = np.unique(segments).size
 
         assert shap_values[0].shape == np.zeros((1, n_segments)).shape
+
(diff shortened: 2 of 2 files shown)


WillemSpek:
I see that the build is failing due to the unit test also asserting that the number of shapeley values should equal that of n_segments.:

`>>> assert shap_values[0].shape == np.zeros((1, n_segments)).shape` 

I believe the unit tests should be changed in similar style to the solution.

dianna/methods/kernelshap.py
```diff
@@ -116,6 +116,7 @@ def explain(
         self.image_segments = self._segment_image(self.input_data, n_segments,
                                                   compactness, sigma,
                                                   **slic_kwargs)
+        n_segments = np.unique(self.image_segments).size
```
cwmeijer, line 119: slic returns an integer mask I agree you should find the set of those to get n_segments.
https://scikit-image.org/docs/stable/api/skimage.segmentation.html#skimage.segmentation.slic

tests/test_kernelshap.py
```diff
@@ -53,7 +53,7 @@ def test_shap_explain_image(self):
         n_segments = 50
```
cwmeijer, line 53: n_segments is explicitly set to 50, and only used to test against. It doesn't make a lot of sense to overwrite this value, as proposed in this PR, before it is tested against. The proposed changes would test whether the output of the method (shap_values and segments) are consistent with each other, which is a good test by itself. However, I think that the idea of this test was to check the outcome of the method against some knowledge that we have prior to running the method, like some ground truth or regression test.
It's a shame it is not explained why this value of 50 is what is expected as an outcome exactly.  @geek-yang, git blame points at you ;-) (https://github.com/dianna-ai/dianna/commit/b8dd755236d2acf3ee3a49420f481402226ea4ac) Could you explain this value and/or review this PR?

cwmeijer (commented):
@yang, can you have a look at this?

tests/test_kernelshap.py
geek-yang: 😅 To be honest Chris I forgot why I set it to be exactly 50. But I think it is a bug given that the number of labels can be slightly different, as `skimage.segmentation.slic` already says `n_segments` is actually "The (approximate) number of labels".

So it should be corrected as what @WillemSpek suggests. Thanks for spotting this @cwmeijer and @WillemSpek . Shame on me 😂

PS: Oh yes, the commit was made at the first week of 2022, right after new year. You guys were all on holiday and I'm the only one working at that moment. What a piece of painful memory 😭. But anyway...my bad 😂 .

dianna/methods/kernelshap.py
geek-yang: Yes, I think I made a mistake and it is a bug here. The actual `n_segments` should be updated as the actual number of slices can be different from the given `n_segments`. Thanks for spotting it @WillemSpek ! 👍

geek-yang (approved):
Thanks for finding and fixing this bug @WillemSpek ! It helps a lot!

yang:
> @yang, can you have a look at this?

Looking great team! Keep cranking!

Issue #633, opened 2023-08-27
Tokenizer sequence length mismatch


WillemSpek:
<h2> Problem Description </h2>
I know this problem was already mentioned in #531 but LIME for text still seems to crash in specific cases with the tokenizer. The problematic characters that I have found thus far are '-', '\' '(' and ')'. 

<h2> Possible Solutions </h2>

<h3> Super Batching </h3>
One way to attack this problem is to group the tokenized sentences for each of their sequence length, i.e. a dictionary and keep track of their indices:

```
super_batch = 
{
  'tokens of length 14': np.ndarray,
  'tokens of length 15': np.ndarray
}
```
 
```
 indices = 
 {
  'tokens of length 14': np.ndarray,
  'tokens of length 15': np.ndarray
 }
 
```
 At the end iterate through each of the batches in the super batch and compute the model scores. Then, reindex these scores to obtain result. The great disadvantage is that the tokenization process is now different than it ought to be and explanation quality might be affected, but we will avoid the tokenization problem.
 
 <h3> Fixing the tokenizer </h3>
 I am not that great of a regex guy :disappointed: but @stefsmeets  was already able to fix a bunch of problems with the tokenizer. Perhaps there is way to fix for all special characters. 
 
 <h3> Unit testing </h3>
 Perhaps we should make a unit test for the tokenizer to challenge it with a bunch of special characters in order to capture its issues for in the future. A unit test for each of the special characters will allow us to attack the problem more effectively.

stefsmeets:
The main issue here is that you are tokenizing an already tokenized sentence. If it is possible to just tokenize once, that would be better and avoid this problem altogether.

The regex solution works around this by adding spaces around `UNKWORDZ`. I'm not great with regex either, but with https://regex101.com/ I can usually figure it out 😅

I already added a bunch of failing strings here:
https://github.com/dianna-ai/dianna/blob/3284afe6aee489afecb3754aba9fad851b66e56f/tests/test_lime.py#L169

It would be useful to at least add the failing strings (maybe mark with [`@pytest.mark.xfail`](https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) with a link to this issue.

PR #637, opened 2023-09-14
added tests for the 'tokenizer problem'

WillemSpek:
Some added tests for the 'tokenizer problem' that I encountered during the work as discussed with @stefsmeets in issue #633 . I have marked the tests with `mark.xfail` as suggested by [the guidelines here](https://docs.pytest.org/en/7.1.x/how-to/skipping.html ) (end of the page).


stefsmeets (approved):
Looks good to me! I didn't know you could mark individual parameters like this 👍

PR #649, opened 2023-10-18
Fix CD/CI notebooks

### LINKED ISSUE #648: Fix notebooks running failures
https://github.com/dianna-ai/dianna/actions/runs/6509287273

geek-yang:
Fix notebooks and build for CI/CD.

This PR closes #648 .


elboyran:
Some checks did not pass. This time dashboard and build for 3.11.

geek-yang:
> Some checks did not pass. This time dashboard and build for 3.11.

Hi @elboyran , thanks for your quick comment. This is not ready for review. Because of our settings in cd/ci, some tests only run when the PR is switched from draft to ready for review, that's why. Sorry for the confusion.

Anyway, I manage to fix all the builds and notebooks ci, but now dashboard fails 😓. I will try to fix it.

geek-yang:
Hi @elboyran @cwmeijer , I manage to fix tests for the notebooks and this PR is ready for review now!

.github/actions/install-python-and-package/action.yml
```diff
@@ -44,7 +44,6 @@ runs:
         python3 -m pip install tensorflow-cpu
 
     - name: Install DIANNA
-      if: steps.cache-python-env.outputs.cache-hit != 'true'
```
geek-yang, line 47: When fixing the CI for notebooks, the build for python 3.11 on macos fails unexpectedly. Error message indicates that `pytest` is not found. I figured out that it was due to the installation, in which pytest was installed for python 3 but the ci calls pytest from python via system path. Therefore, it can be fixed by simply change `run: pytest -v` to `run: python3 -m pytest -v`.

However, this build will fail again if the caching of package installation is used (not sure why...). By removing the cache, it works nicely. But this makes the CD/CI slower than before. If anyone has better solution, please let me know. Thanks.

elboyran: No other solution from me.

elboyran (approved):
Thanks  for fixing it, even if it's slower now.

cwmeijer (approved):
Some interesting things are happening here. I tried changing some things to more expected variations but I couldn't make them pass the builds. Great that you found a setup that indeed does all tests. Merge it!

geek-yang:
> Some interesting things are happening here. I tried changing some things to more expected variations but I couldn't make them pass the builds. Great that you found a setup that indeed does all tests. Merge it!

😅 It took me quite a while to find a solution. Tbh, I'm also not entirely sure about the cause of the failure. But they can be fixed this way. Thanks a lot for the review!

PR #661, opened 2023-11-14
646 Implement LIME for tabular data 

geek-yang:
Add LIME tabular to the methods and provide example notebooks for both regression and classification tasks:

- [x] Add LIME tabular to methods
- [x] Adapt LIME to the dianna API
- [x] Example notebook for classification
- [x] Example notebook for regression
- [x] Add tests

Also fixes the inconsistent input variable names in init (https://github.com/dianna-ai/dianna/blob/add_lime_tabular/dianna/__init__.py).

Note that the method returns an array of importance scores for all input features (which means `num_features` is disabled). For more information about this decision, check this issue #665.


geek-yang:
We need a visualizer for the results. I created an issue for it #662 . Currently the function returns an LIME explainer object.

tutorials/lime_tabular_penguin.ipynb
cwmeijer, line 1: The data_instance loading is done in an unexpected section. It would be better to do this in the data loading section above.

A bit lower, you're using the variable name 'exp' which is short but ambiguous. It would be better to call it 'explanation'.

tutorials/lime_tabular_weather.ipynb
cwmeijer, line 1: Idem to my comments on the penguin notebook.

cwmeijer (changes requested):
Looks great Yang! I still have some requests like additional tests, and also something to discuss maybe: the necessity to pass the train data to the method.
Great work, also on the notebooks!

tests/methods/test_lime_tabular.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_lime_tabular_regression_correct_output_shape(self):
+        """Test the output of explainer."""
+        training_data = np.random.random((10, 2))
+        input_data = np.random.random(2)
+        feature_names = ["feature_1", "feature_2"]
+        exp = dianna.explain_tabular(run_model, input_tabular=input_data, method='lime',
+                             mode ='regression', training_data = training_data,
+                             feature_names=feature_names, class_names=['class_1'])
+
+        assert len(exp) == len(feature_names)
```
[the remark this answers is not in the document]
geek-yang: I like the idea of adding a "naive test" for it. I thought about it. But then I was thinking that we added a test for timeseries mainly because we implement those methods ourselves (referring to some papers and methods like lime segmentation, for instance, but purely implementing those methods mostly from scratch).

 This LIME tabular is entirely from LIME. We simply put a wrapper around it. It is somehow tested in the original implementation of LIME:

https://github.com/marcotcr/lime/blob/master/lime/tests/test_lime_tabular.py

But I do agree that it is nice to add a test for it. Maybe let's create an issue and design a simple test for it?? This PR already gets too big...I would prefer to do it in a separate PR.

dianna/methods/lime_tabular.py
```diff
@@ -0,0 +1,132 @@
...
+        self.mode = mode
+        init_instance_kwargs = utils.get_kwargs_applicable_to_function(
+            LimeTabularExplainer, kwargs)
+
+        # temporary solution for setting num_features and top_labels
+        self.num_features = len(feature_names)
+        self.top_labels = len(class_names)
+
+        self.explainer = LimeTabularExplainer(
+            training_data,
```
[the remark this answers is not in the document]
geek-yang: See my comments above.

dianna/methods/lime_tabular.py
```diff
@@ -0,0 +1,132 @@
...
+            **init_instance_kwargs,
+        )
+
+    def explain(
+        self,
+        model_or_function,
+        input_tabular,
+        labels=(1, ),
+        num_samples=5000,
+        **kwargs,
```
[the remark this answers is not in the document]
geek-yang: Oh yes, well spot! Thanks Chris! Just miss those stuff. I will add them.

tutorials/lime_tabular_weather.ipynb
geek-yang: Done.

tutorials/lime_tabular_penguin.ipynb
geek-yang: Thanks for the suggestions! Done.

cwmeijer (approved):
🏅 🧸 🏆 🦾 🎉

Issue #672, opened 2023-12-11
Fix images in readme table (datasets)


cwmeijer:
Images show as paths

elboyran:
Solved by #673

PR #673, opened 2023-12-11
Update README.md

elboyran:
Testing with one image (logo)


elboyran:
Will fix #672

cwmeijer (approved):
Looks good! No idea why we used to work with "`"s before. It looks like @cpranav93 even added a couple last month while adding timeseries to the table (https://github.com/dianna-ai/dianna/commit/79caaf48fe56178ee56bf609f958fcd341ce905a).
Anyway, it renders correctly now! Thanks!

cpranav93:
Interesting. It appears to be something that is automatically done. Possibly a VSCODE thing?

PR #675, opened 2023-12-12
664 Fix notebooks for CD/CI

geek-yang:
Fix the notebooks in CD/CI.

Also update the target testing version of python to be py310, as py39 is a bit older now.


geek-yang:
Ok. I think I find the solution. It is again due to some versions conflicts, this time about `onnx`. Check this [post/issue](https://github.com/onnx/tensorflow-onnx/issues/2262).

We need to use onnx < 1.15.0.

cpranav93 (approved):
Seems to be working indeed! Good find!

Issue #679, opened 2023-12-13
Update Tutorials README


elboyran:
The tutorials [readme](https://github.com/dianna-ai/dianna/tree/main/tutorials#tutorials), do not include the tabular modality.
Also classification vs. regression.

PR #681, opened 2023-12-15
Update README.md

elboyran:
Solves #679


geek-yang (approved):
Thanks for updating it! Looks great!

cpranav93:
One suggestion:

- To be consistent with other entries in the table, it might good to add a line between the two models as done for time-series.

Otherwise looks great and the links work.

elboyran:
> One suggestion:
> 
>     * To be consistent with other entries in the table, it might good to add a line between the two models as done for time-series.
> 
> 
> Otherwise looks great and the links work.

Hi @cpranav93 , what do you mean exactly? I cannot see a missing line, so probably I'm looking at the wrong place.

cpranav93:
image

Although looking at the code, it seems like the line should be there and it may be a rendering issue on my web browser, so please ignore it if that's the case

elboyran:
> image
> 
> Although looking at the code, it seems like the line should be there and it may be a rendering issue on my web browser, so please ignore it if that's the case

Thanks, you are fast! I was looking at the wrong README- my both PRs were called "Update README" ;-)

elboyran:
> image
> 
> Although looking at the code, it seems like the line should be there and it may be a rendering issue on my web browser, so please ignore it if that's the case

Hmmm, in my branch there is already a line: 
image
image

PR #684, opened 2024-01-05
676 Split tests for modularized code

### LINKED ISSUE #676: Split tests to be consistent with the modularized code
In #543 #544 we split all these methods, but the corresponding tests are not yet split (for instance, see those tests for lime https://github.com/dianna-ai/dianna/blob/main/tests/test_lime.py, tests for images and texts are put in the same file). It is nice to have them separately, to be consistent with the methods.

geek-yang:
Split tests for each method (to be consistent with the modularized code).

This PR closes #676 .


tests/methods/test_lime_text.py
```diff
@@ -165,21 +102,24 @@ def tokenizer():
...
                  marks=pytest.mark.xfail(reason='poor handling of /')),
-    pytest.param('    ', 0,
-                 marks=pytest.mark.xfail(reason='Repeated whitespaces')),
-    pytest.param('I like   whitespaces.', 4,
+    pytest.param(
+        '    ', 0, marks=pytest.mark.xfail(reason='Repeated whitespaces')),
+    pytest.param('I like   whitespaces.',
+                 4,
                  marks=pytest.mark.xfail(reason='Repeated whitespaces')),
 ])
```
cwmeijer, line 122: I don't really see how this is an improvement, but I'm fine with it either way :-)

cwmeijer (approved):
Great! Thanks for splitting! Much nicer structure to build on now. ✨

tests/methods/test_lime_text.py
geek-yang: Those are done by the `pre-commit` command, not by me intentionally 😂 .

PR #688, opened 2024-01-11
Add tabular example to getting start in readme

### LINKED ISSUE #686: Add tabular example in the Getting started section of the README.
Now we have text, image and time series example, but not tabular one (also the classification/regression version?).

geek-yang:
Add tabular example to getting start section in readme.

This PR closes #686 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ dianna.visualization.plot_image(explanation[labels.index(class_a)], original_dat
 model_path = 'your_model.onnx'  # model trained on images
 timeseries_instance = pd.read_csv('your_data_instance.csv').astype(float)
 
-num_features = len(timeseries_instance )  # The number of features to include in the explanation.
+num_features = len(timeseries_instance)  # The number of features to include in the explanation.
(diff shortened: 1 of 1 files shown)


elboyran:
There seems to be a dead link found by the markdown checker.

elboyran (approved):
Looks good to me, thanks!
The only remark I have is for "Run using the XAI methods of your choice.", We don't have RISE supported (yet). right? Maybe we still leave it like you wrote to be future proof.

geek-yang:
> Looks good to me, thanks! The only remark I have is for "Run using the XAI methods of your choice.", We don't have RISE supported (yet). right? Maybe we still leave it like you wrote to be future proof.

Thanks for the review. We have both RISE and SHAP (soon to be merged) for now and plan to support RISE in the future. So I think  we can leave it like this.

geek-yang:
> There seems to be a dead link found by the markdown checker.

It is due to a broken link for the joss badge. Not relevant with this PR, but it is fixed here.

PR #693, opened 2024-01-18
added kwargs tests and code to raise warning for extra kwarg

cpranav93:


cpranav93:
#693 resolves this issue. It is still in a draft stage but the idea is as follows:

- From the list of kwargs passed to the init call, progressively remove kwargs that are being used by: the init function for the explainer, and then the explain function. 
- If any kwargs are leftover, these are assumed to be extra and a warning is currently raised to let the user know that extra kwargs are being passed.

Whether a warning or an error should be passed is to be decided based on whether additional kwargs get passed out of DIANNA to external functions like lime/shap.

cpranav93:
Update:
extra kwargs not accepted by the function are not passed to the function. Therefore we could opt to raise an error if we choose. 

I have now implemented the warnings for each method type and written tests for the same (except tests for tabular). This PR is ready for review

dianna/__init__.py
```diff
@@ -154,4 +171,6 @@ def _get_explainer(method, kwargs, modality):
     method_kwargs = utils.get_kwargs_applicable_to_function(
         method_class.__init__, kwargs
     )
+    for key in method_kwargs.keys():
```
SarahAlidoost, line 174: Please add an inline comment here to explain why this is needed.

tests/test_kwargs.py
```diff
@@ -0,0 +1,166 @@
...
+                                    axis_labels=axis_labels,
+                                    preprocess_function=None,
+                                    top_labels=None,
+                                    num_features=10,
+                                    num_samples=10,
+                                    return_masks=True,
+                                    positive_only=False,
+                                    hide_rest=True,
+                                    extra_kwarg=None
+                                    )
```
SarahAlidoost, line 64: can you also add a test to check the message of warning specifically `{kwargs}` is correct?

SarahAlidoost (commented):
@cpranav93 thanks for the fixes. The implementation works fine. However, I found that there is a function [get_kwargs_applicable_to_function](https://github.com/dianna-ai/dianna/blob/618fcf770e269b66d307a13c47262b4d69522e31/dianna/utils/misc.py#L38). I think the warning should be implemented in this function in `misc` script rather than individual warnings in each method in `init` script. Here is a suggestion for that function:
```python
function_kwargs = {
        key: value
        for key, value in kwargs.items()
        if key in inspect.getfullargspec(function).args
    }

    # check if function is a bound method
    if hasattr(function, '__self__'):
        not_applicable = kwargs.copy()
        for key in function_kwargs:
            not_applicable.pop(key)
        if not_applicable:
            msg = f'Please note the following kwargs are not being used: {not_applicable}'
            warnings.warn(msg)
```
The `test_kwargs` that you implemented still works with this suggestion. I might miss some points regarding the development. If so, please let me know :blush:

cpranav93:
@SarahAlidoost, thanks for the review. Regarding your last point, the issue is that the kwargs are being used by two different functions so the warning should only be raised after the second function's kwargs are also popped. If implemented in the "get_kwargs_applicable_to_function" function, then it would check it after every function and not after both have been called.

dianna/__init__.py
cpranav93: Added an inline comment

tests/test_kwargs.py
cpranav93: Combined the two with assertWarnsRegex in the latest commit.

SarahAlidoost (approved):
@cpranav93 thanks for addressing the comments and explanations. I see the point regarding kwargs. Considering the current implementations of the modules, I agree that your fixes work fine. I approved this PR.

Issue #691, opened 2024-01-17
RISE explanation values


cpranav93:
In the text explainer, rise values are never negative while lime values go to the negative. The lack of a unified range of values makes it difficult to compare across the methods. Should our visualizer scale the output to a fixed range before plotting?

elboyran:
Yes, it should.

cpranav93:
@elboyran do you know if the visualizer does it already or do we need to do implement it?

elboyran:
I don't know it yet. Will look at it as part of my work on issue #663.

PR #703, opened 2024-01-30
Unify outputs types

### LINKED ISSUE #640: unify the function outputs in the API
Create an overview in this issue with outputs changed etc.
Consider using an output parser that can be reused and tested

SarahAlidoost:
closes #640 
relates #691, #713, #711

See an overview of changes [here](https://github.com/dianna-ai/dianna/issues/640#issuecomment-1916752378).


dianna/__init__.py
```diff
@@ -71,7 +71,7 @@ def explain_image(model_or_function, input_image, method, labels, **kwargs):
         kwargs: These keyword parameters are passed on
 
     Returns:
-        One heatmap (2D array) per class.
+        An array (np.ndarray) containing the heat maps for each class.
```
cwmeijer, line 74: Nice and thorough!
It would be even better if we could have type hints in the function signature.
We once decided (about a year ago) to start using type hints. We decided back then to not change the complete code at once, but only update code wherever we are working on it already. Now that you already created such thorough docstrings including input and output types, it shouldn't be such a big task to add those to the function signature as well:
def explain_image(model_or_function:Union(str, Callable), input_image:ndarray, method:string, labels:Iterable[int], **kwargs) -> ndarray:
If that's too much effort or is too troublesome in certain cases, only the output types would already help.
Let me know if you disagree to include this in this PR.

dianna/methods/kernelshap_image.py
```diff
@@ -100,7 +104,7 @@ def explain(
         https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.slic
 
         Returns:
-            Explanation heatmap of shapley values for each class (np.ndarray).
+            Explanation heatmap for each class (np.ndarray).
```
cwmeijer, line 107: I think I would prefer the old comment:
```suggestion
            Explanation heatmap of Shapley values for each class (np.ndarray).
```

dianna/__init__.py
SarahAlidoost: Good idea. I added types wherever possible. It would be nice if `mypy` is also added to the workflows to check the typing. see issue https://github.com/dianna-ai/dianna/issues/85

SarahAlidoost:
> Thanks for your PR! Clearly, a lot of work went into it! I only reviewed 1 or 2 files yet. Could you have a look at my comment on type hints? I also see that the notebooks are failing. image I think it's because you changed the output type of some explainer so it doesn't match the notebook's code anymore. Could you try and fix those?

Thanks for reviewing. I addressed your comment about typing. Regarding the notebook error, as discussed, they will be fixed  in a separate PR, see #711

dianna/methods/kernelshap_image.py
```diff
@@ -123,12 +127,27 @@ def explain(
...
-        return shap_values, self.image_segments
+        # create heat_maps where shape is (n_classes, *image_segments.shape)
+        n_classes = len(shap_values)
+        heat_maps = np.zeros((n_classes, *self.image_segments.shape))
+
+        # fill the heat_maps with shap values for each class and segment
+        for i, shap_value in enumerate(shap_values):
+            class_heat_map = heat_maps[i]
+            for index in self.image_segments.flat:
+                class_heat_map[self.image_segments == index] = shap_value[0][index - 1]
```
cwmeijer, line 144: I'm a bit confused about what is in shap_values. It seems there is a list in there for every class. Then inside that list there is another level (hence '[0]') and only in there is the actual list of shap_values for that class. Except for that extra level, it makes sense to me, but maybe we can help the reader of the code a bit with more descriptive variable name?
Maybe shap_value_lists instead of shap_values to imply that there is a multilevel thing in there for instance?

dianna/methods/kernelshap_image.py
```diff
@@ -123,12 +127,27 @@ def explain(
...
+            heat_maps[i] = class_heat_map
```
cwmeijer, line 145: You could consider extracting these lines into its own method and maybe even create a test for it.
Both are just suggestions, doing neither of these things at this point is totally fine with me.

cwmeijer (approved):
Thanks for this PR! Great addition to our code base and improvement in usability!
I made some suggestions, but all are to some extend details. Whether you change anything or not, I approve merging this PR.
🎖

elboyran:
Some checks should pass before we merge.

elboyran:
Yes, you are correct.

dianna/methods/kernelshap_image.py
SarahAlidoost: Thanks, I agree. It is renamed now.

dianna/methods/kernelshap_image.py
SarahAlidoost: I understand that the code will be more readable if this code block goes to its own function. However, I think it does not need to be a method of the class as the `heatmaps` or `image_segments` are already returned. Instead of a method, I refactored the code block as a function in this module.

SarahAlidoost:
> Thanks for this PR! Great addition to our code base and improvement in usability! I made some suggestions, but all are to some extend details. Whether you change anything or not, I approve merging this PR. 🎖

Thank you for reviewing and nice comments :+1:

elboyran:
When do we merge this PR?

PR #728, opened 2024-03-19
Fix plots

### LINKED ISSUE #206: Improve explanation results in the tutorial notebooks of RISE text

### LINKED ISSUE #691: RISE explanation values

### LINKED ISSUE #696: Fix text tutorial relevace maps visuation
The highlighted text (relevance scores) in **all** text tutorials is not visible in HTML
E.g. [RISE on text](https://github.com/dianna-ai/dianna/blob/main/tutorials/rise_text.ipynb).

### LINKED ISSUE #713: Fix visualization functions
There are several problems related to the visualization functions:
1- The value ranges of plot, see issue #691
2- Some functions create double plots by `ax.imshow` and` plt.show()` and `return fig`
3- lack of colorbar

issues of text visualization: 
1- text visualization does not support if a word is repeated because it uses `get()` method of a dictionary to map a word to an importance value, see [here](https://github.com/dianna-ai/dianna/blob/f4736d4746e4651100139a9f6ec9ca5487c94627/dianna/visualization/text.py#L34-L40)
2- text visualization takes an argument `input_tokens` that can be optional
3- see #696 and #206 and #716

see related issues of text visualization #302, #435, #317, #285

### LINKED ISSUE #716: check why html is not always showing up in notebooks
maybe because of github viewer or something similar

SarahAlidoost:
closes #713 

and 
closes #691 
closes #696 
closes #206 
closes #716 

I think this pull request will close all issues listed above. Reviewer please check it. Thanks :blush:


SarahAlidoost:
@cwmeijer this pull request is ready for a review. I noticed an issue with text LIME in the dashboard, but it is not related to this pull request, see #738

dianna/visualization/image.py
```diff
@@ -62,9 +68,11 @@ def plot_image(heatmap,
                    labelbottom=False,
                    labelright=False,
                    labeltop=False)
-    if show_plot:
-        plt.show()
+
```
elboyran, line 71: If I understand correctly the logic you'd like to always show the plot by default (that's why `show_plot` is `True` by default). Then in the code, it's a bit strange that the only use of the parameter is when it is **not** true. Isn't therefore better for readability `hide_plot` (with default `False`) a better name? Then the `if` statement will read more logically. Just a suggestion.

dianna/visualization/tabular.py
```diff
@@ -32,21 +32,24 @@ def plot_tabular(
...
+    fig, ax = plt.subplots()
     colors = ["r" if x >= 0 else "b" for x in top_values]
     ax.barh(top_features, top_values, color=colors)
     ax.set_xlabel(x_label)
     ax.set_ylabel(y_label)
 
-    if show_plot:
-        plt.show()
+    if not show_plot:
+        plt.close()
```
elboyran, line 50: See the same comment I left in `image.py`, line 71.

dianna/visualization/text.py
```diff
@@ -1,69 +1,101 @@
...
+    # add colorbar
+    sm = plt.cm.ScalarMappable(
+        cmap=plt.get_cmap(colormap),
+        norm=plt.Normalize(vmin, vmax)
+        )
+    sm.set_array([])
+    fig.colorbar(sm, ax=ax, orientation='horizontal', aspect=20, use_gridspec=True)
+    # TODO add alpha to the colorbar
+
+    if not show_plot:
```
elboyran, line 85: See the same comment I left in `image.py`, line 71.

dianna/visualization/text.py
```diff
@@ -1,69 +1,101 @@
...
+        plt.close()
+
+    if output_filename:
+        plt.savefig(output_filename)
+
+    return fig, ax
+
+
+def _get_text_color(importance, vmin, vmax, colormap, alpha):
+    if importance is None:
```
elboyran, line 95: An inline comment on what `importance` is would help here. It's possible to retrieve, but one needs to look the the parent function's body. Not suggesting a full parameter description as that would be redundant.

dianna/visualization/timeseries.py
```diff
@@ -49,19 +51,21 @@ def plot_timeseries(
         current_ax.set_ylabel(y_label_current)
         current_ax.label_outer()
 
-    _draw_segments(axs, cmap, segments)
+    _draw_segments(axs, cmap, segments, heatmap_range)
+
+    if not show_plot:
```
elboyran, line 56: See the same comment I left in `image.py`, line 71.

tests/test_dashboard.py
```diff
@@ -8,11 +8,22 @@
     pip install pytest-playwright
```
elboyran, line 8: I'm not an expert on `streamlit`. Maybe, could ask @stefsmeets  for a quick look at your suggested changes.

elboyran (commented):
Good work. Have some minor comments.

tests/test_dashboard.py
stefsmeets: Happy to have a look, let me know which bits.

SarahAlidoost: Thanks @elboyran for the comment. the dashboard is created using streamlit, see [here](https://github.com/dianna-ai/dianna/tree/main/dianna/dashboard). 
@stefsmeets can you please have a look at the updated docstring [here](https://github.com/dianna-ai/dianna/pull/728/files/10a9ba169446c1813dd60ef601b9107360982ec1#diff-37f37eddbedf6e2deac58b85ba731696800ae8670a251d57d6cc5bde95e8baa1L1-L16)? Together with @cwmeijer, I found that the old docstring is not complete, the  `playwright codegen http://localhost:8501` returns error because first the server should be started.  So I updated the docstring saying to use `streamlit` instead of `playwright` to run tests.

stefsmeets (approved):

elboyran:
[like]  Elena Ranguelova reacted to your message:
________________________________

dianna/visualization/image.py
SarahAlidoost: @elboyran I agree with you. When I was refactoring this part, I was also thinking in the same way. I noticed that this interface of visualization functions i.e. `show_plot` as an input argument has already been used in many places e.g. code, tests, notebooks, .... Renaming it here adds a lot of changes to this pull request that was not initially intended. If you agree, I skip renaming it in here, but I submit an issue for it.

dianna/visualization/text.py
SarahAlidoost: I added docstring to this function.

SarahAlidoost:
> @SarahAlidoost , good work! Have some minor comments.

Thanks for reviewing, I addressed your comments. If the PR looks okay, can you approve it?

dianna/visualization/image.py
cwmeijer: I understand your reasoning. I'm not sure it's actually better (maybe only slightly) of the presence of another parameter, output_filename, also there to output something. Also hide_plot sounds a bit like a negative, which would make hide_plot=False make a double negative. I'm sort of fine with either option but I would suggest not spend time on this and just leave it as it is.

elboyran: OK.

elboyran (approved):
Nice!

Issue #720, opened 2024-02-29
Update DIANNA README with the FRB usecase data and model


elboyran:
Sections 

- [x] ONNX models - [Time-series](https://github.com/dianna-ai/dianna/tree/main?tab=readme-ov-file#time-series-1)

loostrum:
Fixed in #733

PR #733, opened 2024-03-25
Add FRB model to main readme

loostrum:
Closes issue #720 

I've added a link to the FRB model and paper describing its generation.

I've not added a link to the dataset as it isn't public. We just use a single sample from the dataset in our tutorial.


loostrum:
I've also fixed a broken link found by the markdown link checker, unrelated to this PR. Not sure how we missed that earlier but at least it's fixed now.

elboyran (approved):

PR #744, opened 2024-04-09
add a link to the time series masking blog post from docstrings

### LINKED ISSUE #564: Add documentation/docstrings/examples of strategic masking based on wave function
Referring to the comments in PR #554, we implement very strategic and smart masking approach based on wave functions for both images and timeseries, for instance:

for images
https://github.com/dianna-ai/dianna/blob/3f5b388ab0faa36876730a87e18f7b2876621d35/dianna/methods/rise.py#L245

for timeseries
https://github.com/dianna-ai/dianna/blob/3f97c304ad16fd4f3adce528405632a414f207b9/dianna/utils/maskers.py#L20-L22

But since the method is a bit tricky and abstract, it is not so easy to understand. Currently we don't have it documented in detail. We should add explicit documentation/docstrings/examples to explain our approach.

cwmeijer:
Document the masking strategy by linking to the blog post I wrote about it.
Fixes #564 .
Note that the first commit are the actual planned changes. But I also changed a function name there to make it a bit more clear, which wasn't supposed to be part of this PR, but broke some imports. The second commit is fixing those imports.

### DIFF
--- a/dianna/methods/lime_timeseries.py
+++ b/dianna/methods/lime_timeseries.py
@@ -4,7 +4,7 @@ from fastdtw import fastdtw
 from lime import explanation
 from lime import lime_base
 from dianna import utils
-from dianna.utils.maskers import generate_masks
+from dianna.utils.maskers import generate_time_series_masks
 from dianna.utils.maskers import mask_data
 from dianna.utils.predict import make_predictions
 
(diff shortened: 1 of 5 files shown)


elboyran (approved):
Nice. Looks neat and link to the blog post.

Issue #739, opened 2024-03-27
Kernel SHAP tabular tutorial exampels are missing from the READMEs


elboyran:
Add them to:

- [x] Tutorials [README](https://github.com/dianna-ai/dianna/tree/main/tutorials#tutorials) table
- [x] Main README [checks table](https://github.com/dianna-ai/dianna?tab=readme-ov-file#tutorials)

It is the best to do it after resolving the #697 issues.

PR #745, opened 2024-04-10
Update READMEs with KernalSHAP mark and links

elboyran:
Updated Tutorials section of main and the Tutorials README including the KernaSHAP tutorials on tabular data.

Solves #739


cwmeijer (dismissed):
Looks good!

cwmeijer (approved):
Not sure why the notebooks are failing on this branch but as it's just a couple of md changes It can't be the reason. Looks good btw!

PR #752, opened 2024-04-17
Add test for rise timeseries on onnx file

### LINKED ISSUE #690: explainer does not work with onnx model file path
The explaine_timeseries (for example) mentions: 
model_or_function (callable or str): The function that runs the model to be explained _or_
                                             the path to a ONNX model on disk.

However, passing the path to ONNX model on disk does not work as the model output it usually double and we need it to be float. We should address this either by converting it ourselves, making the model work with double, or add it as a requirement on the model.

loostrum:
Closes #690 

#690 describes the explainer not working with an onnx file for timeseries. After some testing, I found that the season prediction model requires float32 input, while numpy/python use float64 by default. This makes the season prediction model crash when using the onnx file. I don't think this should be considered a bug in dianna, rather the user should convert the data using the preprocess function. I added a test that does this.

### DIFF
--- a/tests/methods/test_rise_timeseries.py
+++ b/tests/methods/test_rise_timeseries.py
@@ -12,10 +12,34 @@ def test_rise_timeseries_correct_output_shape():
     input_data = np.random.random((10, 1))
     labels = [1]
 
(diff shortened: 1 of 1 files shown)


elboyran (approved):
Clear.

PR #769, opened 2024-05-22
732-overview-tutorial

elboyran:

PR #772, opened 2024-05-29
Fix data file paths for Colab

### LINKED ISSUE #750: text vector file not found in lime_text on colab
It seems the files are not to be found on the same (relative) paths on colab as they are when running from the repo. We should look for the files in the repo on the colab storage, or we should make such files available through zenodo or directly from our repo.
image

loostrum:
Use files from dianna installation folder in all tutorials.
    
This simplifies the colab setup as we no longer need to download data
files.

The relative paths throughout the tutorials were replaced by paths
starting from the dianna install dir, which should work on all
platforms.

The indentation in the colab setup was also fixed, this was 2 spaces
instead of 4 in many places.

Fixes #750

### DIFF
--- a/dianna/dashboard/_movie_model.py
+++ b/dianna/dashboard/_movie_model.py
@@ -1,6 +1,6 @@
 import os
 import numpy as np
-from _shared import label_directory
+from _shared import data_directory
 from scipy.special import expit as sigmoid
 from torchtext.vocab import Vectors
 from dianna import utils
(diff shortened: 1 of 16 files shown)


elboyran (approved):
Ah, nice work, thanks! 
Will do the same in the new (renamed demo) overview tutorial in my PR #769 .

elboyran:
Sorry, I have noticed only now that the `root_dir` was introduced, but not used in the [Kernelshap weather regression tutorial](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/KernelSHAP/kernelshap_tabular_weather.ipynb).

PR #774, opened 2024-05-29
Move docs logo to sidebar

loostrum:
Old situation:

image

New situation:

image

Unfortunately the logo is a png with white background, so it doesn't look as nice in the sidebar. @elboyran do you have a vector version of the logo? Or one with transparent background?


elboyran:
Unfortunately, I do not have one with a transparent background (I've created it long ago just using Word fonts and symbols...).  I've tried online BG removal tools, but the result does not look so good. Would we know the exact color of the sidebar?

elboyran (approved):
Nice, that's the proper solution, indeed.

loostrum:
Using the sidebar background color would be a good solution! I found the color in the Sphinx RTD theme CSS, it's hex [2980B9](https://www.color-hex.com/color/2980b9)

loostrum:
Blue background version:

image

PR #778, opened 2024-05-31
Update MANIFEST.in

elboyran:
With the current installation, the code reorganization was not reflected in the installation. 
Now all the ` data`, `labels`, and `models` folders have been included explicitly.


loostrum (approved):

loostrum (approved):
In the future I would recommend that we try the Zenodo download option for everything, so we don't have to include these files in the install at all. But for now, it is important that we fix the notebooks and the changes look good, so let's merge it!

PR #792, opened 2024-06-07
783 simplify overview tutorial

elboyran:
Make the overview tutorial better.


SarahAlidoost (approved):
@elboyran thanks, looks very good :+1:  I was able to run the notebook. One minor comment: is there a reason why we don't plot the results of kernelshap, the last method in the notebook?

elboyran:
> @elboyran thanks, looks very good 👍 I was able to run the notebook. One minor comment: is there a reason why we don't plot the results of kernelshap, the last method in the notebook?

@SarahAlidoost thanks for reviewing! If you recall during the last stand-up (or maybe you were not there?) there was the opinion that I should skip the whole **Explainers** section as it was redundant. Still, I opted for keeping it and gradually "simplifying" it, which includes firstly elaborating on LIME (corresponding to a missing LIME for Binary MNIST tutorial), then less for the second explainer and not showing the last visualization. The point of the section for me is:

- Stress again we support a few explainers
- Show it's as easy as specifying the method's name (and some hyperparameters)
- Finish the overview by triggering the interest and "advertising" all combinations of modalities-methods and pointing to the dashboard (where the methods' comparison) is the most obvious advantage.

I hope this PR achieves that.

PR #776, opened 2024-05-29
Overview tutorial and other tutorial changes

elboyran:
Fixes #732, #755 and #770

PR #796, opened 2024-06-12
Automatic data downloading with pooch

loostrum:
A downloader utility was added, which contains an overview of all data files we use, hosted on either Github or Zenodo.
With the `dianna.utils.downloader.download` function, one can download and use any of these files. The files are downloaded to an OS cache directory by default, and the checksums are verified so any changes in the online files are detected.

The tutorials were updated to use this functionality.

While rerunning the notebooks I noticed the colormaps were mostly wrong (default viridis instead of bwr). This is due to https://github.com/dianna-ai/dianna/pull/776, where the colormaps were removed from the notebooks, but the defaults were not set in the dianna code for all modalities (only in the dashboard it seems). In this PR I added the bwr default to the visualization module where not already present.


tests/test_downloader.py
cwmeijer, line 1: Nice and thorough! This is quite a heavy test. Could we run this for only one build config for instance instead of all build configurations?

dianna/visualization/image.py
```diff
@@ -10,13 +10,14 @@ def _determine_vmax(max_data_value):
...
 
 
-def plot_image(heatmap,
-               original_data=None,
-               heatmap_cmap=None,
-               heatmap_range=(None, None),  # (vmin, vmax)
-               data_cmap=None,
-               show_plot=True,
-               output_filename=None):
+def plot_image(
```
cwmeijer, line 13: why this reformatting? just some autoformatter I suppose? it makes the pr a bit larger.

cwmeijer (approved):
Looks great! I tried one notebook and run the tests. Awesome! 🍕🍔🍟

elboyran:
@loostrum if al models are downloaded from Zenood now, why to keep the `models` folder? Or at least to remove the non-needed files?

PR #802, opened 2024-06-18
enable passing of external Axes to plot on

egpbos:
Adds to the plot_image and plot_tabular functions an optional argument ax: plt.Axes. When given, this is the Axes that will be used to plot on and the internal plt.subplots call is skipped. This is useful for using these functions in custom multi-panel plots (we want to use this in a paper).

We actually only need this for the `plot_image` function, but I added it to `plot_tabular` as a bonus ;) The `plot_timeseries` function is too complicated, I don't have time to add it there right now, but the principle will be similar (except now the ax has to be passed through three layers of helper functions).

### DIFF
--- a/dianna/visualization/image.py
+++ b/dianna/visualization/image.py
@@ -1,4 +1,6 @@
+from typing import Optional
 import matplotlib.pyplot as plt
+import numpy as np
 
 
 def _determine_vmax(max_data_value):
@@ -10,14 +12,15 @@ def _determine_vmax(max_data_value):
     return vmax
 
 
-def plot_image(heatmap,
-               original_data=None,
+def plot_image(heatmap: np.ndarray,
+               original_data: Optional[np.ndarray] = None,
                heatmap_cmap='bwr',
                heatmap_range=(None, None),  # (vmin, vmax)
                data_cmap=None,
-               show_plot=True,
+               show_plot: bool = True,
                output_filename=None,
-               ax=None):
+               ax: Optional[plt.Axes] = None,
+) -> plt.Figure:
     """Plots a heatmap image.
 
     Optionally, the heatmap (typically a saliency map of an explainer) can be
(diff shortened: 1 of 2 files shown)


egpbos:
P.S.: "we" here is me and @cwmeijer :)

egpbos:
Whoops, forgot to rebase on main, done now.

stefsmeets (approved):
Looks good to me! 🚀🚀

Minor nitpick would be to also add types to the heatmap plot while you are at it 😅

egpbos:
Your nitpick is my command ;)

I did not hint it completely. The matplotlib stuff is just too complex and bloaty and increases maintenance cost. I think the reference in the docstring to go look at mpl.Axes.imshow is best for those options.

egpbos:
The CI failures are HTTP timeouts unrelated to this PR.

dianna/visualization/image.py
```diff
@@ -38,13 +42,18 @@ def plot_image(heatmap,
         show_plot: Shows plot if true (for testing or writing plots to disk
                    instead).
         output_filename: Name of the file to save the plot to (optional).
+        ax: matplotlib.Axes object to plot on (optional).
 
     Returns:
         None
```
cwmeijer, line 48:
```suggestion
        fig: figure
        ax: axes
```

dianna/visualization/image.py
```diff
@@ -10,13 +12,15 @@ def _determine_vmax(max_data_value):
...
+               original_data: Optional[np.ndarray] = None,
                heatmap_cmap=None,
                heatmap_range=(None, None),  # (vmin, vmax)
                data_cmap=None,
-               show_plot=True,
-               output_filename=None):
+               show_plot: bool = True,
+               output_filename=None,
+               ax: Optional[plt.Axes] = None,
+) -> plt.Figure:
```
cwmeijer, line 23:
```suggestion
) -> (plt.Figure, plt.Axes):
```

cwmeijer (approved):
Top! Could you first accept my two other suggestions (same as in the other file)? Thanks for this addition!

egpbos:
Will you merge when you think it's done?

PR #814, opened 2024-06-26
Dashboard test failing

laurasootes:
This PR fixes the error that made the dashboard test fail. It was a small error in the page titles.

Additionally, the positioning of the dianna image in the sidebar is fixed. Somehow it overlapped with the sidebar pages


elboyran (approved):
Good fix!

PR #825, opened 2024-07-26
support ArrayLike data in to_xarray

egpbos:
This commit makes dianna swallow more kinds of data items without users having to mangle them manually themselves. As long as your data is numpy ArrayLike it will go. The only blocker for this was the direct use of the .ndim attribute on data, which assumes it is already a numpy array rather than something that can trivially be converted to a numpy array. A PIL.Image is an example of such an ArrayLike type. We use this in explainable_embedding to feed items into the OpenAI CLIP model.

I sprinkled in some nice type hinting, of course, but not for the other parameters, because they would become monstrosities.

### DIFF
--- a/dianna/utils/misc.py
+++ b/dianna/utils/misc.py
@@ -1,6 +1,8 @@
 import inspect
 import warnings
 from pathlib import Path
+import numpy as np
+import numpy.typing
 
 
 def get_function(model_or_function, preprocess_function=None):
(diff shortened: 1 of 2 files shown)


egpbos:
Fixed the import issue from the previous CI build (tested locally). However, now we have a different, unrelated issue that we have to wait to get fixed (see #827 and #829).

loostrum:
I rebased on main to get the linter fix in here. The changes themselves look good, nice generalization of to_xarray! Linter's happy as well, so let's merge.

loostrum (approved):

elboyran:
@egpbos many checks are not passing.

egpbos:
Yep, those are mentioned above. They are unrelated to this PR.

cwmeijer:
closing and reopening for rerunning tests

PR #827, opened 2024-07-29
use tensorflow_probability[tf] extra to match versions

egpbos:
Fixes #826.

Related to #434 and possibly #727.

PR #828, opened 2024-07-29
fix ruff invocation in linting workflow

egpbos:
Title says it all. The linting workflow is broken, this fixes it (I hope).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,27 +22,7 @@ extras = dev
 """
 
 [tool.ruff]
-exclude = [
-    ".bzr",
-    ".direnv",
-    ".eggs",
-    ".git",
-    ".hg",
-    ".mypy_cache",
-    ".nox",
-    ".pants.d",
-    ".ruff_cache",
-    ".svn",
-    ".tox",
-    ".venv",
-    "__pypackages__",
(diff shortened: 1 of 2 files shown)


egpbos:
Note that the torchtext failures seem unrelated as mentioned in #827. The only workflow that matters is the one that actually does work again now: the linting one :) https://github.com/dianna-ai/dianna/actions/runs/10145802048/job/28052379417?pr=828

pyproject.toml
```diff
@@ -22,6 +22,35 @@ extras = dev
...
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "venv",
+    ".venv",
+    "scripts",
+]
+
```
stefsmeets: Most of these are already in the defaults, I recommend to use `extend-exclude` instead (https://docs.astral.sh/ruff/settings/#extend-exclude):

```suggestion
extend-exclude = [
    ".venv",
    "scripts",
]

```

stefsmeets (approved):
Looks good to me! 🚀

egpbos:
`.venv` is also in default exclude it seems, so also removed that in the force pushed amended commit.

Issue #816, opened 2024-07-03
Dashboard: add text


laurasootes:
Add some more text to make the use of the dashboard and the examples a bit clearer. Specifically:

- [x] Landing page: rename “pages”, and clarify what images/text/timeseries means in this context
- [x] Landing page: what is the dashboard
- [x] Page titles: make page titles more informative than simple “Image explanation"
- Examples: explain what the example is: what is the data, what is the task, what do the colours mean Image classification page:
- [x] Image classification
- [x] Tabular classification
- [x] Tabular regression
- [x]  Text classification
- [x] Time series classification weather
- [x] Time series classification FRB

Suggestion from Sarah to remove the pages section fully

elboyran:
Adding this comment in the first seemingly relevant still open issue I found:

- [ ] fix typos and reformulate a bit.
![image](https://github.com/user-attachments/assets/982c9ede-68e8-4cd7-9949-d237103755c7)

- [ ] edit to avoid repetitions
![image](https://github.com/user-attachments/assets/4434b47e-679a-4519-b51f-c368e19384f5)

elboyran:
For the colors of the attribution maps, borrow/get inspired by this nice [explanation](https://arxiv.org/pdf/2206.03208v1#subsubsection.A.3.1).

elboyran:
Landing page implemented: 
![image](https://github.com/user-attachments/assets/7b1b4210-1eb3-4042-81c1-09fc77e10d05)

elboyran:
Image page and example:

![image](https://github.com/user-attachments/assets/918f0fa7-c289-484b-8df9-e3315e770193)
-----------------------------------------------
![image](https://github.com/user-attachments/assets/094f226b-0587-4051-9964-da67195d33d6)

elboyran:
Tabular page and examples:

![image](https://github.com/user-attachments/assets/c97027cc-3dd2-4a44-997c-c72864149910)
------------------------------------------------
![image](https://github.com/user-attachments/assets/c5f847ab-e9e0-4822-969f-6ec7601699b4)
--------------------------------------------------
![image](https://github.com/user-attachments/assets/63ad42e6-5f85-4532-a595-7e92a55f57de)

elboyran:
Text page and examples:

![image](https://github.com/user-attachments/assets/b0dc16a5-e0a6-4b5e-a561-0426822370a4)
--------------------------------------

[...]

PR #830, opened 2024-07-30
817 dashboard move menu

### LINKED ISSUE #817: Dashboard: move menu
Move the menu (home, images, text, time series links) from the left to the top

laurasootes:
This PR moves the menu from the left to the top of the page. This deviates from streamlits defaults, and requires the use of dynamic page imports. Also the ‘original menu’ has to be hidden in the settings.

These adjustments cause the changed parameters in each page to be conserved between pages. To solve this I added page-specific keys to the variable elements, which are reset when you switch pages.

Also added icons to the menu items :)

Solves #817


dianna/dashboard/Home.py
```diff
@@ -1,6 +1,8 @@
 import streamlit as st
 from _shared import add_sidebar_logo
 from _shared import data_directory
+from streamlit_option_menu import option_menu
+import importlib
```
SarahAlidoost:
```suggestion
import importlib
import streamlit as st
from _shared import add_sidebar_logo
from _shared import data_directory
from streamlit_option_menu import option_menu
```

SarahAlidoost: Here I just moved the `importlib` to the top to make `isort` happy. 

Can you please add `streamlit_option_menu` to `dashboard` in [setup.cfg](https://github.com/dianna-ai/dianna/blob/430be09d51c26cfc5fc7ed9b87126cb7965a544b/setup.cfg#L90)?

dianna/dashboard/Home.py
```diff
@@ -16,26 +18,64 @@
...
+    It wraps carefully selected XAI methods in a simple, uniform interface. It's built by,
+    with and for (academic) researchers and research software engineers working on machine
+    learning projects.
 
+    ### Pages
 
-### More information
+    - <a href="/Images" target="_parent">Images</a>
+    - <a href="/Text" target="_parent">Text</a>
+    - <a href="/Time_series" target="_parent">Time series</a>
```
SarahAlidoost, line 55: Do we still need this list? I know removing this isn't a part of issue #817, but perhaps we can also remove them here.

dianna/dashboard/Home.py
```diff
@@ -16,26 +18,64 @@
...
+    "Images": "pages.Images",
+    "Text": "pages.Text",
+    "Time series": "pages.Time_series"
+}
 
-st.image(str(data_directory / 'logo.png'))
+# Set up the top menu
+selected = option_menu(
+    menu_title=None,
+    options=["Home", "Images", "Text", "Time series"],
```
SarahAlidoost:
```suggestion
    options=list(pages.keys()),
```

SarahAlidoost (changes requested):
@laurasootes nice work :+1: I liked the small icons next to each menu option. I was able to run the dashboard, but some of the tests in [test_dashbord](https://github.com/dianna-ai/dianna/blob/main/tests/test_dashboard.py) fail on [`expect(page).to_have_title(...)`](https://github.com/dianna-ai/dianna/blob/430be09d51c26cfc5fc7ed9b87126cb7965a544b/tests/test_dashboard.py#L93).  and also some minor comments on changed files.

dianna/dashboard/Home.py
laurasootes: Fair point, I think it is best to address this in [#816](https://github.com/dianna-ai/dianna/issues/816)

laurasootes:
@SarahAlidoost thanks for your review! I implemented your code suggestions and the linting now passes. 

I looked at the error you mention for the dashboard titles. I do not know why they would arise. Also I do not get them locally and don’t see them in the logs in the github actions either so I am a bit puzzled. Could you maybe try yo run them again?

I added your comment for the pages section on the home page to #816, because I think there will be some more discussion on what should and should not be on the home page

SarahAlidoost:
> @SarahAlidoost thanks for your review! I implemented your code suggestions and the linting now passes.
> 
> I looked at the error you mention for the dashboard titles. I do not know why they would arise. Also I do not get them locally and don’t see them in the logs in the github actions either so I am a bit puzzled. Could you maybe try yo run them again?
> 

You are right. Perhaps something related to running the tests locally on my side. I ran the dashboard test as `pytest -v -m dashboard --dashboard` on the main branch too and got the same error. Somehow the titles of pages are for example `"Text · Streamlit"` instead `"Text"` on my side :smile: . As you don't get the errors locally, I approved this PR. The failed GitHub actions are due to issue #837 .

SarahAlidoost (approved):

Issue #818, opened 2024-07-03
Dashboard: modifu input data section for multiple examples


laurasootes:
instead of “load example data” make selection boxes in the “input data” section for each specific example

<img width="669" alt="image" src="https://github.com/dianna-ai/dianna/assets/95367983/e9ab2393-fdf9-4ca7-b388-536bf91e7d99">

PR #832, opened 2024-07-31
790 add scientific use case frb to dashboard

### LINKED ISSUE #790: Add scientific use case (FRB)  to dashboard
Start from the [FRB RISE tutorial](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/RISE/rise_timeseries_frb.ipynb).

### LINKED ISSUE #839: use new download functionality in dashboard
see notebooks for example
it's in `utils/downloader.py`

laurasootes:
This PR adds the FRB use case to the dashboard.

The FRB case and data is quite different from the weather example, and therefore requires a lot of FRB-specific data conversions and such, making the code not that pretty. It only works for RISE (as is the case in the notebook). Since it is not straightforward, I made an issue to get it to also work for LIME #836 . 
Fixes #790 

Additionally, I split the left menu; it now starts with a choice for example or own data, and only then shows the choices/inputs required depending on example/own data. I think this makes the dashboard less confising and less visually overwhelming: you do not see choices and buttons you do not need. Extension to #818 

I updated the tests in accordance with the new input data layout

Also, added the DIANNA.utils.downloader for the data used in the dashboard. Fixes #839


laurasootes:
@SarahAlidoost  would you perhaps have time to review this PR?

dianna/dashboard/pages/Images.py
```diff
@@ -11,6 +11,8 @@
 from _shared import data_directory
 from _shared import label_directory
 from _shared import model_directory
+from _shared import reset_method
+from _shared import reset_example
```
SarahAlidoost:
```suggestion
from _shared import reset_example
from _shared import reset_method
```

SarahAlidoost: to fix isort error

dianna/dashboard/pages/Text.py
```diff
@@ -10,6 +10,8 @@
 from _shared import add_sidebar_logo
 from _shared import label_directory
 from _shared import model_directory
+from _shared import reset_method
+from _shared import reset_example
```
SarahAlidoost:
```suggestion
from _shared import reset_example
from _shared import reset_method
```

dianna/dashboard/pages/Time_series.py
```diff
@@ -6,12 +7,15 @@
 from _shared import _get_method_params
 from _shared import _get_top_indices_and_labels
 from _shared import _methods_checkboxes
-from _shared import add_sidebar_logo
+from _shared import  add_sidebar_logo
```
SarahAlidoost:
```suggestion
from _shared import add_sidebar_logo
```

dianna/dashboard/pages/Time_series.py
```diff
@@ -51,25 +60,75 @@
         contributing positively towards the classification decision are
         indicated in red and those who contribute negatively in blue.
         """)
+    elif load_example == "Scientific case: FRB":
+        ts_file = (data_directory / 'FRB211024.npy')
+        ts_model_file = (model_directory /
+                        'apertif_frb_dynamic_spectrum_model.onnx')
```
SarahAlidoost: The frb model `apertif_frb_dynamic_spectrum_model.onnx` does not exist in the [`model_directory`](https://github.com/dianna-ai/dianna/tree/790-add-scientific-use-case-frb-to-dashboard/dianna/models). Can you please add the model? Later, when the downloader functionality is implemented, this part can be refactored.

dianna/dashboard/pages/Time_series.py
```diff
@@ -51,25 +60,75 @@
...
+        ts_label_file = (label_directory / 'apertif_frb_classes.txt')
+
+        # FRB data must be preprocessed
+        def preprocess(data):
+            """Preprocessing function for FRB use case to get the data in the right shape."""
+            return np.transpose(data, (0, 2, 1))[..., None].astype(np.float32)
+
+        ts_data = open_timeseries(ts_file)
+        ts_data_dianna = ts_data.T[None, ...]
+        ts_data_model = ts_data[None, ..., None]
```
SarahAlidoost: The variable names `ts_data_model` and `ts_data_dianna` are not informative, because the only difference between them is their shape; one is the transpose of the other. Could you rename these variables to something more descriptive?  Also, please add an inline comment explaining the purpose of defining these two variables.

dianna/dashboard/pages/Time_series.py
```diff
@@ -51,25 +60,75 @@
...
+
 
 if not (ts_file and ts_model_file and ts_label_file):
     st.info('Add your input data in the left panel to continue')
     st.stop()
 
-ts_data = open_timeseries(ts_file)
+if load_example != "Scientific case: FRB":
+    ts_data_model = open_timeseries(ts_file)
+    ts_data_dianna = ts_data_model
```
SarahAlidoost: the same comment as above, the names `ts_data_model` and `ts_data_dianna` are not informative, could you rename these variables to something more descriptive?  Also, please add an inline comment explaining the purpose of defining these two variables.

tests/test_dashboard.py
```diff
@@ -175,14 +188,15 @@ def test_timeseries_page(page: Page):
...
 
-    page.locator('label').filter(
-        has_text='Load weather example').locator('span').click()
+    page.locator("label").filter(has_text="Use an example").locator("div").nth(1).click()
+    expect(page.get_by_text("Select an example in the left")).to_be_visible()
+    expect(page.get_by_text("Weather")).to_be_visible()
+    expect(page.get_by_text("FRB")).to_be_visible()
 
-    expect(page.get_by_text('Select a method to continue')).to_be_visible()
+    page.locator("label").filter(has_text="Weather").locator("div").nth(1).click()
```
SarahAlidoost, line 200: @laurasootes can you also add a test for when frb example is selected?

SarahAlidoost (changes requested):
@laurasootes nice work! :+1: I like the new interface with separate sections in the left menu, much cleaner now. I was able to run the dashboard, and its tests. Just a few comments on the changed file. Also, I noticed that the quality of frb images is not the same as those in the notebook. can you check it? see here:

from dashboard 
frb_dashboard

from notebook
frb_nb

dianna/dashboard/pages/Time_series.py
laurasootes: I renamed them to ts_data_predictor and ts_data_explainer respectively and added a comment to make it clearer

dianna/dashboard/pages/Time_series.py
laurasootes: I renamed them to ts_data_predictor and ts_data_explainer respectively and added a comment to make it clearer

dianna/dashboard/pages/Time_series.py
laurasootes: Thanks for pointing this out. I implemented now the download function, also for the other example models. Then that issue will also be resolved in this PR

laurasootes:
@SarahAlidoost Thanks for your review!
I have implemented your suggestions:
- for the FRB data, I used the downloader function now, and also implemented these for the other examples
- for the naming of the variables related to the FRB adjustments; I renamed them and added some inline comments as you suggested. I hope the code is clearer now

As for the result; the difference is caused by the parameters used. Changing the model parameters gives me this:

image

(The difference in darkness is caused by the overlay, which was not done in the notebook)

dianna/dashboard/pages/Time_series.py
```diff
@@ -88,16 +147,24 @@
...
             with st.spinner(f'Running {method}'):
-                explanation = func(serialized_model, ts_data=ts_data, **kwargs)
+                explanation = func(serialized_model, ts_data=ts_data_explainer, **kwargs)
 
-            segments = _convert_to_segments(explanation)
+            if load_example == "Scientific case: FRB":
+                # normalize FRB data and get rid of last dimension
+                fig, _ = plot_image(explanation[0, :, ::-1].T,
+                                    ((ts_data + np.min(ts_data)) / (np.max(ts_data) + np.min(ts_data)))[::-1]
+                                    )
```
SarahAlidoost: Is the normalization done to scale the range of `ts_data` to [0,1]? if so, the equation should be `(value - min_value) / (max_value - min_value)`

dianna/dashboard/pages/Time_series.py
```diff
@@ -88,16 +147,24 @@
...
         with col:
             with st.spinner(f'Running {method}'):
-                explanation = func(serialized_model, ts_data=ts_data, **kwargs)
+                explanation = func(serialized_model, ts_data=ts_data_explainer, **kwargs)
 
-            segments = _convert_to_segments(explanation)
+            if load_example == "Scientific case: FRB":
+                # normalize FRB data and get rid of last dimension
+                fig, _ = plot_image(explanation[0, :, ::-1].T,
+                                    ((ts_data + np.min(ts_data)) / (np.max(ts_data) + np.min(ts_data)))[::-1]
```
SarahAlidoost:
```suggestion
                                    ((ts_data - np.min(ts_data)) / (np.max(ts_data) - np.min(ts_data)))[::-1]
```

dianna/dashboard/pages/Time_series.py
SarahAlidoost: one more comment: plotting the original data in this example doesn't contribute any new information and only makes the figure darker. I suggest removing it.

SarahAlidoost:
> As for the result; the difference is caused by the parameters used. Changing the model parameters gives me this:

@laurasootes I amnot sure if the difference is caused by the parameters. Here is the screenshot of my dashboard with the same parameters as yours. There are two main differences: 
1. the figure's aspect ratio: yours is square, while mine is rectangular. I checked the `explanation[0, :, ::-1].T.shape` that is `(32, 64)`. 
2. the x and y axis labels are missing in my figure. To check this, I saved the fig returned by `plot_image()` as `fig.savefig('...')` or in the notebook after `4 - Plot the explanation and input data next to each other` visualization.plot_image(heatmap) returns the same plot as mine. Can you please check these two issues?

frb_dashboard

laurasootes:
> > As for the result; the difference is caused by the parameters used. Changing the model parameters gives me this:
> 
> @laurasootes I amnot sure if the difference is caused by the parameters. Here is the screenshot of my dashboard with the same parameters as yours. There are two main differences:
> 
> 1. the figure's aspect ratio: yours is square, while mine is rectangular. I checked the `explanation[0, :, ::-1].T.shape` that is `(32, 64)`.
> 2. the x and y axis labels are missing in my figure. To check this, I saved the fig returned by `plot_image()` as `fig.savefig('...')`. Can you please check these two issues?
> 
> frb_dashboard

To compare with the one in the notebook I temporarily fixed the aspect radio and added the axis labels in the dashboard (and removed the background original image), so that I could check how far the red part of the explanation reaches as the ’shape’ of the explanation was my concern. (I did not add the axis labels in the dashboard as I think it does not add any information for the explanation). Besides these formatting things I believe your result is the same as mine right?

SarahAlidoost:
> > > As for the result; the difference is caused by the parameters used. Changing the model parameters gives me this:
> > 
> > 
> > @laurasootes I amnot sure if the difference is caused by the parameters. Here is the screenshot of my dashboard with the same parameters as yours. There are two main differences:
> > 
> > 1. the figure's aspect ratio: yours is square, while mine is rectangular. I checked the `explanation[0, :, ::-1].T.shape` that is `(32, 64)`.
> > 2. the x and y axis labels are missing in my figure. To check this, I saved the fig returned by `plot_image()` as `fig.savefig('...')`. Can you please check these two issues?
> > 
> > frb_dashboard
> 
> To compare with the one in the notebook I temporarily fixed the aspect radio and added the axis labels in the dashboard (and removed the background original image), so that I could check how far the red part of the explanation reaches as the ’shape’ of the explanation was my concern. (I did not add the axis labels in the dashboard as I think it does not add any information for the explanation). Besides these formatting things I believe your result is the same as mine right?

thanks for the explanations :+1: , now I understand what causes the differences. Yes, if I add the aspect ratio and the lables, the figures are the same. :smile:

laurasootes:
@SarahAlidoost thanks for your suggestions. I removed the background from the plot, which also removed the normalisation problem :)

SarahAlidoost (approved):
@laurasootes thanks for addressing the comments :+1:

PR #840, opened 2024-08-13
Fix docstring of visualization tabular

### LINKED ISSUE #822: Visualization: `plot_tabular` does not work for x as `np.ndarray`
The input `x` in `plot_tabular` is defined as `Array of feature importance scores`. However, it returns an error if `x` is `np.ndarray`. There isnot a test for this in` test_visualization.py`. You can reproduce the error as:

```python
import numpy as np
def test_plot_tabular_with_array():
    """Test plot tabular data."""
    x = np.random.rand(5, 3)
    y = [f"Feature {i}" for i in range(x.shape[1])]
    fig, ax = plot_tabular(x=x, y=y, show_plot=False)
    assert fig is not None
```

the error
```bash
>       top_values = [x for _, x in sorted(zip(abs_values, x), reverse=True)][:num_features]
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

SarahAlidoost:
closes #822

### DIFF
--- a/dianna/visualization/tabular.py
+++ b/dianna/visualization/tabular.py
@@ -18,7 +18,7 @@ def plot_tabular(
     """Plot feature importance with segments highlighted.
 
     Args:
-        x (np.ndarray): Array of feature importance scores
+        x (np.ndarray): 1D array of feature importance scores of one instance
         y (List[str]): List of feature names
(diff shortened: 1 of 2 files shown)


cwmeijer (approved):
Great!

Issue #789, opened 2024-06-06
Add tabular example in dashboard


elboyran:
As a start use the [LIME](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/LIME/lime_tabular_penguin.ipynb) and [KernelSHAP](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/KernelSHAP/kernelshap_tabular_penguin.ipynb) tutorials on the penguin data (classification task).

Brainstorm oh can we add also the regression examples [LIME](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/LIME/lime_tabular_weather.ipynb) and [KernelSHAP](https://github.com/dianna-ai/dianna/blob/main/tutorials/explainers/KernelSHAP/kernelshap_tabular_weather.ipynb) on the weather datatset?

cwmeijer:
Because the weather dataset was already demoed in the time series tab, we chose to go with the penguins dataset.

Issue #801, opened 2024-06-14
in lime_tabular, the input argument `feature_names` cannot be None


SarahAlidoost:
in lime_tabular, `None` is set as a default value for the input argument `feature_names`, see [here](https://github.com/dianna-ai/dianna/blob/48a9994ac4cb652b4c6e54217d6bf0cf9df1e62d/dianna/methods/lime_tabular.py#L17). Later, the variable `self.num_features` is calculated as `len(feature_names)` which does not work on NoneType data, see [here](https://github.com/dianna-ai/dianna/blob/48a9994ac4cb652b4c6e54217d6bf0cf9df1e62d/dianna/methods/lime_tabular.py#L65). Note in [lime](https://lime-ml.readthedocs.io/en/latest/lime.html#lime.lime_tabular.LimeTabularExplainer.explain_instance), `self.num_features` is not optional.

The same issue for the input argument `class_names` with the default value `None`.

For both input arguments, the default values should not be `None` or the `None` case should be handled inside the function.

PR #844, opened 2024-08-22
838 create tabular tab to dashboard and redesign loaded data results #819

### LINKED ISSUE #523: Move collapsable XAI method parameters in method blocks
Looks good!
One thing is the XAI method's parameters. Those are method specific and I suggest we therefore move them in the method's (output) section like in my mockups above.

_Originally posted by @cwmeijer in https://github.com/dianna-ai/dianna/issues/454#issuecomment-1411821343_

### LINKED ISSUE #819: Dashboard: redesign loaded data results
- [x] create one clear box containing method section checkboxes, and number of results to show.
- [x] To the right of that add the predicted class
- [x] add the “click to modify parameters” option per individual method
- [x] For the expanded modify parameters, include them per emthod. See also #523 
- [x] add axis labels (e.g. “class”)
- [x] add explanation text for the example at the top of the page, under the title. See also #816 
- [x] move “select method to continue” above method selection bar

image

### LINKED ISSUE #838: Create tabular tab to dashboard
add new tabular tab to dashboard

laurasootes:
This PR adds a tabular page to the dashboard (#838 ) and finalises all dashboard redesigns (#819 ). 

Tabular page:
Page is implemented similar to the other pages (text, images, time series).
- Other than the other pages, the tabular page requires the user to upload the training data, which has to be provided as npy file. 
- For this page it is not obligatory to upload a label text file, because you can also use it for regression models, in which case labels are not necessary. If labels are not uploaded, the page will treat the example as a regression model.
- The available methods are currently RISE and LIME.
- For the usage, a user uploads their tabular data csv file (full file). The data is read as pandas dataframe and the full table is shown in the dashboard. The user then clicks on a row in the table, which is the data for which the prediction and explanation is done.
- A simple test to check the basic functionality of the page is added to the dashboard tests.
- Note that a tabular example will be added in #789 

Dashboard redesign changes:
- _get_method_params is converted into get_params, such that method parameter seleciton is per method
- For all pages a streamlit container (basically a box) is created at the top which includes the predicted class, number of top classes to show, method selection and method parameters. The different components are only shown when relevant.
- function _get_top_indices_and_labels is adjusted such that for a regression model the predicted value is printed in the dashboard instead of the predicted class

Other changes:
- naming of keys for streamlit elements is done in a prettier way

fixes #838 and fixes #819 and fixes #523


dianna/methods/lime_tabular.py
```diff
@@ -63,6 +63,7 @@ def __init__(
             LimeTabularExplainer, kwargs)
 
         # temporary solution for setting num_features and top_labels
+        # when fixed, also fix in dashboard Tabular.py -> _feature_names
```
elboyran, line 66: Is it fixed in the next PR?

elboyran (approved):

dianna/methods/lime_tabular.py
laurasootes: No, the temporary fix itself was already here, it has been added in december [PR 661](https://github.com/dianna-ai/dianna/pull/661/files). I did not try to fix the general problem

laurasootes: I believe [issue 801](https://github.com/dianna-ai/dianna/issues/801) mentions the general issue

Issue #829, opened 2024-07-29
torchtext end-of-life and broken


egpbos:
As noted in #827 and other recent PRs with breaking CI workflows, the torchtext package seems to be breaking down. It is no longer being developed (see https://github.com/pytorch/text/issues/2250).

Some options:

1. Find a workaround ourselves.
2. Look for a fork that **is** still maintained and switch to that.
3. Replace torchtext as a dependency.

Option 3 seems the most attractive to me, naively, but I haven't looked deeply into how unique the functionality is that we use. We only use torchtext in two ways:

- `from torchtext.data import get_tokenizer` in `utils/tokenizer.py`
- `from torchtext.vocab import Vectors` in `test/utils.py`, in a couple of notebooks and in the dashboard.

Can these easily be replaced? If not, a fourth option presents itself:

4. Cannibalize torchtext for these parts only.

loostrum:
Yikes, I guess we should have followed up on the deprecation warning earlier ;) 
I'll see if we can replace torchtext. In the linked pytorch issue they also suggest taking from the torchtext source what you need, so option 4. may not be the worst idea, we can just put them in a dianna text utils file.

cwmeijer:
[Spacy](https://github.com/explosion/spaCy), a dependency of DIANNA already, may provide both functionalities. Looking at their code, it seems so. I came across a `vocab.Vectors` and a `get_tokenizer` in their code. That would be a nice solution I think (that's an option 3 for ya 😉).

SarahAlidoost:
As a temporary fix to get CI green again, torch version is pinned in #841. Once this issue is resolved, the temporary fix can be removed.

egpbos:
We may need to take another look at a more future-proof solution for this, because we may be running against the limits of the pinned versions we're using now. E.g. [this notebook run](https://github.com/dianna-ai/dianna/actions/runs/25170854839/job/73790264787?pr=879) may be failing due to outdated circular dependencies between `pytorch_lightning` and other torch stuff (not sure yet, but even if this is not the issue in this particular case, it's probably still going to be causing issues if we don't upgrade torch beyond 2.3.1 now).

There are a couple of references to this issue both in commit texts and in code (e.g. in the install python action yml file), so for clarity I'm reopening this one instead of creating a new issue.

egpbos:
I didn't realize the torchtext dependency itself was already gone since #853, nice! That means we can drop the workaround version pins again. Working on this in #883.

[...]

PR #853, opened 2024-09-18
Removes dependency on deprecated torchtext (refs #829)

cwmeijer:
See #829.
This PR gets rid of:
- [x] uses of `torchtext.get_tokenizer`
- [x] uses of `torchtext.vocab.Vector`


pyproject.toml
```diff
@@ -69,11 +66,10 @@ unfixable = []
 
 per-file-ignores = {}
 
-
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
-[tool.ruff.lint.isort]
```
loostrum, line 76: I'm assuming this is because of some linter deprecation/error?

loostrum (changes requested):
The changes look good! Could you add pandas as a dependency to setup.cfg? Then it's good to go.

pyproject.toml
cwmeijer: indeed. Apparently this changed in the last half year and my precommit wouldn't even let me commit anything without the change ;-)

cwmeijer:
> The changes look good! Could you add pandas as a dependency to setup.cfg? Then it's good to go.

I added the dependency already for notebooks, but it was necessary for the dashboard and unit tests, too. I think strictly it shouldn't be necessary for DIANNA base install, but I'm not a 100% sure. As it is necessary for any other type of install, I made pandas a dependency for dianna base.

loostrum (approved):

PR #871, opened 2024-12-04
Update kernelshap_tabular_land_atmosphere.ipynb: increase timeout

cwmeijer:
because timeout occurred for a cell which caused the notebook test to fail

### DIFF
--- a/tutorials/explainers/KernelSHAP/kernelshap_tabular_land_atmosphere.ipynb
+++ b/tutorials/explainers/KernelSHAP/kernelshap_tabular_land_atmosphere.ipynb
@@ -1182,6 +1182,9 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.11.9"
+  },
+  "execution": {
+   "timeout": 1800
   }
  },
  "nbformat": 4,
--- a/tutorials/explainers/KernelSHAP/kernelshap_tabular_land_atmosphere.ipynb
+++ b/tutorials/explainers/KernelSHAP/kernelshap_tabular_land_atmosphere.ipynb
@@ -720,7 +720,7 @@
   },
   {
    "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


cwmeijer:
After increasing the time limit from 300s to 1800s, the same cell still times out. The cell contains the call to dianna in the land_atmosphere notebook. The notebook was running fine before (with 300s limit). We should investigate more.

SarahAlidoost:
> After increasing the time limit from 300s to 1800s, the same cell still times out. The cell contains the call to dianna in the land_atmosphere notebook. The notebook was running fine before (with 300s limit). We should investigate more.

There are two cells in the notebook that take a few minutes to run due to large data. These cells are not suitable for execution with GitHub Action runners. Currently, the github workflow takes about 30 minutes, which is too long to be practical as a check for pull requests. The notebook workflow only runs `pytest --nbmake`, and the results are not used elsewhere. For the notebook `kernelshap_tabular_land_atmosphere.ipynb`, I have added an option to skip running time-consuming cells online. I suggest implementing the same thing for other **tutorial** notebooks if they also contain cells that require more runtime.

SarahAlidoost:
@cwmeijer running notebook `lime_text_eulaw.ipynb` has failed, see action [here](https://github.com/dianna-ai/dianna/actions/runs/12413822969/job/34656532391?pr=871). The problem is new version of scikit-learn (1.6.x), see related issues https://github.com/dmlc/xgboost/pull/11021 and https://github.com/scikit-learn/scikit-learn/issues/30479.

egpbos:
Wow, I missed this PR, but ran into the same issues. I "fixed" all of this in #877, the KernelSHAP notebook is now simply skipped in CI, and the LIME EU law one had another fix in #879 (https://github.com/dianna-ai/dianna/pull/879/commits/82b27cfbe8beb3052ac38a8a4664d3d43bb331a2).

PR #874, opened 2025-02-03
add support for python 3.12

### LINKED ISSUE #873: add support python 3.12
we should investigate if we can add support for 3.12. It's been out long enough.

cwmeijer:
This PR should fix #873.

My progress so far:
- To support 3.12, we had to drop the restriction of TF as only newer versions support 3.12.
- To support newer versions of TF we had to drop our restrictions for onnx.
 
Current status however: 3 build tests fail.

My approach for both dependency conflicts was to remove the problematic restrictions completely. Maybe we should instead only loosen the restrictions to make the tests pass again. I haven't looked into that yet. For now, I'll leave this PR here for anyone who feels like diving into this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
 
       - uses: ./.github/actions/install-python-and-package
         with:
-          python-version: '3.11'
+          python-version: '3.12'
 
       - name: Run unit tests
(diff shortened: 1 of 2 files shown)


egpbos:
I created a new PR branching off this work in #877.

egpbos:
Fixed by #877.

PR #892, opened 2026-06-27
fix release action

egpbos:
Updated the pypa/gh-action-pypi-publish action to use the release version ("v1" no longer exists).

Also removes the explicit __token__ user, because it is the default, no need to specify anymore.

Note: this PR does not activate trusted publishing, which is the recommended mode nowadays (https://github.com/marketplace/actions/pypi-publish), but I don't have access to set it up on PyPI.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.7.0
+current_version = 1.8.0
 
 [comment]
 comment = The contents of this file cannot be merged with that of setup.cfg until https://github.com/c4urself/bump2version/issues/185 is resolved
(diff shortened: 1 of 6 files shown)


egpbos:
Manually triggered the (test) release action: https://github.com/dianna-ai/dianna/actions/runs/28296448919

egpbos:
Setup trusted publishing on PyPI and test PyPI. Restarted the release job: https://github.com/dianna-ai/dianna/actions/runs/28296448919

egpbos:
Ok, the release action worked on test PyPI, but it released 1.7.0. Turns out I didn't actually do the release properly, because the code itself still thinks it's 1.7.0 (setup.cfg at least), d'oh! Will fix that too now and then redo the release.

PR #894, opened 2026-06-29
Fix click CI failure

### LINKED ISSUE #893: CI failing (missing click)
For some reason click is now missing in a bunch of unit tests on CI, making them fail. Probably it was a transitive dependency before and has now been dropped (although on Windows in 3.12 tests are still passing, weird).

egpbos:
Fixes #893. Also replace deprecated Numpy `row_stack` with `vstack` (which is simply an alias, so no change in behaviour).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,7 @@ install_requires =
 [options.extras_require]
 dev =
     bump2version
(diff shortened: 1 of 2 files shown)


loostrum (approved):
LGTM

Issue #878, opened 2026-04-30
dashboard test failures


egpbos:
The [dashboard CI workflow](https://github.com/dianna-ai/dianna/actions/runs/25160257752/job/73752829311?pr=877) is giving a lot of errors. Copilot had a look at it and produced the following summary of what should happen to fix the tests. I don't know enough about streamlit and the dashboard code to assess this in detail, but may have a look later.

# Copilot's analysis

The job is failing in the **dashboard Playwright E2E tests** (workflow: `.github/workflows/build.yml` @ `675fe804bd25b6714b762abf18420bf0bf90d039`, job step: `pytest -v -m dashboard --dashboard`). The failures are all **UI timing/selector brittleness** plus one **incorrect page title expectation**.

Below are focused fixes you can apply in the repo.

---

## 1) Fix the deterministic failure: wrong expected title on Time Series page

Log shows:

- Expected: `Time_series`
- Actual: `Time series`

In `tests/test_dashboard_time_series.py` @ `675fe804bd25b6714b762abf18420bf0bf90d039` line 79:

```py
expect(page).to_have_title('Time_series')
```

Change to what the app actually renders:

```diff
- expect(page).to_have_title('Time_series')
+ expect(page).to_have_title('Time series')
```

(Alternatively, if the intended title is `Time_series`, fix the Streamlit page title in the dashboard code—but based on the test/logs, updating the test is the quickest and most direct fix.)

---

### Root cause
Several tests do:

- `page.get_by_text('Running...').wait_for(state='detached')`

…but Streamlit apps often re-render multiple times; “Running…” may disappear briefly and reappear, and some widgets/outputs render *after* it detaches. That’s consistent with these failures:

- `test_dashboard_image.py`: expects heading `"1"` etc. but element never found
- `test_dashboard_tabular.py`: click timeouts for expander toggle / gridcell inside iframe

### Practical fix: add a stable “app ready” wait helper and use it consistently

Create a small helper (e.g. `tests/dashboard_helpers.py`):

```py
# tests/dashboard_helpers.py
from playwright.sync_api import Page, expect

def wait_streamlit_ready(page: Page, timeout: int = 200_000) -> None:
    # Prefer "networkidle" to let initial page settle
    page.wait_for_load_state("networkidle", timeout=timeout)

    # Streamlit sometimes shows "Running..." during reruns; wait until it's gone
    # Use "last" because there can be multiple occurrences
    running = page.get_by_text("Running...").last
    try:
[...]

PR #896, opened 2026-09-21
Fix dashboard tests, pin streamlit version

v1kko:
- Fixes the broken streamlit tests
- Pins the streamlit version to avoid this in the future


v1kko:
Closes #878, now permanently with a streamlit version pin

PR #898, opened 2026-09-23
Vectorize KernelSHAP image heatmap creation and masking, Fix off-by-one error

v1kko:
Replace the Python loops in `_create_heatemaps` and `KERNELSHAPImage._mask_image` with numpy indexing. Outputs are bit-identical to the loop versions; the originals are kept in tests/methods/test_kernelshap_image_vectorized.py as references for equivalence tests.

Speedup (best of repeated runs, 2 classes, 500 mask samples):

  28x28x1 random image, 81 segments
    _create_heatemaps: 3.7 ms   -> 0.018 ms  (~200x)
    _mask_image:       116 ms   -> 1.4 ms    (~84x)
  224x224x3 astronaut image, 58 segments
    _create_heatemaps: 2293 ms  -> 0.8 ms    (~2800x)
    _mask_image:       3182 ms  -> 663 ms    (~5x, memory-bound)

## Fixed off-by-one error 

SLIC labels segments 1..K while features are 0..K-1. _mask_image mapped
segment k to feature k, whereas _create_heatmaps maps it to feature k - 1,
so feature 0 masked nothing, the last segment was never masked, and each
heatmap value was drawn one segment away from the region that was perturbed.
Index with segmentation - 1 in _mask_image; this also drops the padding
column that was only needed for the out-of-range label.

## Fixed typo

Rename _create_heatemaps to _create_heatmaps


v1kko:
Closes #899

